PR #81, opened 2016-02-02
Update README.md: Fix english grammar (capitalization rule)

fxedel:
The title says almost everything :smiley:

In English, you don't capitalize nouns like in German. So, it's "the **a**pp" instead of "the **A**pp" (German: "die **A**nwendung")


stefan-niedermann:
Thank you. I really want to know why the have capitalized characters at all if they only need if a new sentence starts...
Well, yeah, i know that rule but if you are always writing in german.... :)

Would be interest how a native speaker thinks about german capitalization :D

Issue #91, opened 2016-02-17
Log in screen display misleading error message on wrong URL


marxjohnson:
Source: F-Droid

Version: v0.7.0

Device: Nexus 7 2013

Steps to reproduce:
-     Open the app
-     If not the first run, open Settings
-     Enter an incorrect URL, but the correct username and password
-     Press "Connect"
-     A toast message is displayed, "Wrong username or password"
-     This message should indicate that the URL is incorrect.

fxedel:
You're right. As you can see [here](https://github.com/stefan-niedermann/OwnCloud-Notes/blob/master/app/src/main/java/it/niedermann/owncloud/notes/android/activity/SettingsActivity.java#L242), the app simply uses one string to indicate an invalid login. I've opened #94 and added another string that says: That login is invalid. That's fewer information, but false information is worse.

IMO, the desirable solution would be to mark the input fields directly if they're wrong. That fit's into the [material design spec](https://www.google.com/design/spec/patterns/errors.html#errors-usage). So you'll also have to check if it's a problem with address, username or password.

stefan-niedermann:
Will be fixed in v0.9.0. (Different error messages for different errors)

PR #94, opened 2016-02-20
Quick fix for #91: Use another string to indicate invalid login

fxedel:
A more detailed info would be desirable, but this also works and doesn't tell the user wrong information.


stefan-niedermann:
Thanks for contribution - yeah you are totally right about that.
I got much feedback about this misleading error message. this should be handled better in the future.

PR #98, opened 2016-03-03
Updated FR translation

ageru:
Translation of error_invalid_login
Amends to 2 other translations
Removed some empty lines


stefan-niedermann:
thx

PR #99, opened 2016-03-05
Update russian translations

proninyaroslav:


fxedel:
Thanks for your contributions :)

But as I can see, you've not only edited the string files but also some .idea/* files. Is that intended? If so, could you please explain what you did?

proninyaroslav:
.idea of files I didn't modify. It is made Android Studio :)

fxedel:
Well, then have a look at the [changed files](https://github.com/stefan-niedermann/OwnCloud-Notes/pull/99/files). I think you have to unstage a few files ;-)

proninyaroslav:
Ok, I unstage changes for this files.

proninyaroslav:
I think it is necessary to forbid changes for this files, as Android Studio alter them without the user's knowledge.

stefan-niedermann:
Thank you for contribution.
Yeah, the .idea-Files should be read-only maybe (Could this cause problems when working at the app with idea on other environements?)

- [ ] Clear the not needed changes (.idea etc)
- [ ] Mention yourself as translator not only in strings.xml but on any language file

PR #120, opened 2016-06-26
Improved presentation of errors if login or synchronization fails

### LINKED ISSUE #119: Show an error message if synchronization fails

* the login fails,

korelstar:
This is a first step in order to improve the presentation of errors, if login or synchronization fails. It partly closes #119.


stefan-niedermann:
great, thanks :)

Issue #117, opened 2016-06-06
Notes created or edited in airplane mode are deleted when online


wobblywob:
Steps to recreate:
1. Turn on airplane mode
2. Create or edit notes in the app. The notes appear last as duplicates to the original notes
3. Turn off airplane mode
4. Refresh app. Notes and edits are lost

owncloud server version: 9.0.2
notes version: 2.0.1
android ownCloud Notes: 0.8.1

moshpirit:
Just happened to me. It was a huge note.

Android 6.0, OwnCloud notes 0.8.1 (F-droid)

emunicio:
It happened to me also. However I was not in airplane mode. I didn't have internet connection, and when refreshing after create the node, the note was deleted.

stefan-niedermann:
yep. sorry guys, can confirm this is a bug that didn't appear in 0.7.2. i am very sorry for the delay, i have currently not much time to work on this project. i will fox this as soon as possible.

stefan-niedermann:
ps: pull requests are very welcome :)

korelstar:
Here are the changes from 0.7.2 to 0.8.1: https://github.com/stefan-niedermann/OwnCloud-Notes/compare/ef331d05a0d47b74498c4767d04f7a275096b009...0b8907c34abb170ff791f075e6cbc53e9e449114

I would say, the source for this bug could be in https://github.com/stefan-niedermann/OwnCloud-Notes/blob/0b8907c34abb170ff791f075e6cbc53e9e449114/app/src/main/java/it/niedermann/owncloud/notes/android/activity/NotesListViewActivity.java#L84 : here, `db.getNoteServerSyncHelper().downloadNotes()` is called without synchronizing the notes before. A solution could be to change this call into `db.getNoteServerSyncHelper().synchronize()`. But I haven't tested it yet.

korelstar:
Okay, I had a more detailed look into the source code and found that this is not trivially fixed. There are many (error) cases, which are not considered in the current implementation. I suggest to revise and restructure `NoteSQLiteOpenHelper` and `NoteServerSyncHelper` in order to be more robust.

Some ideas:
- separate database and synchronisation methods, i.e. remove `NoteSQLiteOpenHelper.XAndSync()` methods and set correct `status` of `Note` in remaining methods. This distinction will also be realized in the frontend.
[...]

PR #121, opened 2016-06-29
Show icon if a note is not synchronized (hint to a possible error)

### LINKED ISSUE #119: Show an error message if synchronization fails

korelstar:
This pull request introduces a new icon which is shown on the right margin of a `Note` in the `NoteListViewActivity` if (and only if) the note is not synchronized, i.e., the note is new or was edited and the respective data is only local and not yet on the server side. This gives a hint to check, if there is a problem with the synchronization (partly closes #119). Until now, if synchronization fails, one can't realize this and somebody may think everything is fine.

The implementation introduces the new model `DBNote` which extends `Note` with the `status` attribute in order to display the status of the note. The icons come from the original android settings app: https://github.com/android/platform_packages_apps_settings


stefan-niedermann:
hm, okay, i'll have to check and test this locally, could take some time, sorry.
PS: could you please use the material icon instead of the holo icon for consistent user experience? thanks. (search https://materialdesignicons.com/ for "sync" and you'll find it)

korelstar:
Oh, that's a nice website. Didn't know it. I've updated the icons now.
I'm using this branch as base to work on #117 so it would be nice if it is merged until I'm done (this will still take some evenings) :-)

korelstar:
Okay, my fix for #117 is nearly finished, so all changes done here will be in the next pull request. So I close this one, since the refactoring done here was not perfect. However, comments are welcome and feedback will be integrated into the upcoming pull request.

Issue #66, opened 2016-01-30
Handle remote note deleted


stefan-niedermann:
This is a follow-up-ticket for #65.
# Situation:
1. User starts editing a note on his mobile phone.
2. User deletes note via web interface without leaving edit-mode on mobile
3. User causes autosave through leaving the edit-mode or further editing
4. How should this be handled?
   - **Delete local** note
   - **Recreate remote** note
   - Let **user choose**

cc: @jancborchardt

jancborchardt:
I’d say let’s err on the side of caution here – recreating the remote note seems to be the best course of action. It’s minimally invasive and the scenario is also pretty edge-casey.

korelstar:
With #123, the recreation is implemented (like before?). Hence, this can be closed(?)

PR #123, opened 2016-07-10
Fixing sychronization and concurrency issues

### LINKED ISSUE #76: Editing saves on every keypress. It hammers my connection.
It appears that the application tries to mimic the approach on the ownCloud web front end, which has no explicit save mechanism.

As far as I can see, for every keystroke the whole file is being sent to the server. If let us say the user is editing a largish note on low bandwidth, this becomes a bit of a problem. It also totally screws up my Activity feed.

Why not:
* Have an explicit save mechanism (e.g., an icon, menu option) so the user can choose when to save.
* Save automatically on leaving the note (going back to notes list, dismissing application)
* Save automatically after a period of inactivity. Let us say, if the user has not typed anything in the last *x* seconds.

The options above need not be mutually exclusive. Do note however that any automatic saves should have an undo option, to allow the user to recover from accidents (the web front end provides this via file versioning).

### LINKED ISSUE #104: don't autosave if the note isn't modified
Sometimes I like to copy the raw marckdown code and paste in other app. OCN thinks that I edited the note and save it showing it the first if the list when I didn't.

It would be nice an option to change the sort or cancel the automatic save in these situations.

### LINKED ISSUE #117: Notes created or edited in airplane mode are deleted when online

### LINKED ISSUE #119: Show an error message if synchronization fails

korelstar:
This is the current state of my attempt to fix #117. It's not fully finished yet, but the majority of the work is done. Therefore, please have a look at this pull request and test it with all the test cases you can imagine. I will update the pull request continually.

#### Finished Tasks
- [x] **Database:** new column `REMOTE_ID` (fixes conflicts between parallel creation of different notes locally and remotely)
- [x] **DBStatus:** `LOCAL_CREATED` is not used anymore (distinction if Note is new or edited can be made with checking `REMOTE_ID=0`; fixes conflicts on synchronization and parallel local edits)
- [x] **Note:** model is split in to `DBNote` (representing the local DB entry of a note) and `OwnCloudNote` (representing the remote note, used by `NotesClient`).
- [x] **ItemAdapter/note item layout:** show icon for unsynchronized changes (simplifies debugging, includes all changes done in #121, closes #119)
- [x] **NoteSQLiteOpenHelper:** database upgrade (add the new column and adapts the data without flushing the database)
- [x] **NoteSQLiteOpenHelper:** add more constraints for database updates (fixes lost updates due to parallel changes of notes (from synchronization and user interface))
- [x] **NoteSQLiteOpenHelper:** `updateNoteAndSync()`: only make database changes, if the content really changed (fixes #104, at least the original intention of the author).
- [x] **NoteSQLiteOpenHelper:** remove recurrent code (improves code quality)
- [x] **NoteServerSyncHelper:** only update if network is connected (fixes lots of thrown exceptions)
- [x] **NoteServerSyncHelper:** schedule synchronization requests (if synchronization is active or network not connected; fixes missing push of *all* outstanding changes)
- [x] **NoteServerSyncHelper:** restructure the communication between synchronization task and user interface.
- [x] **NoteServerSyncHelper:** consolidate synchronization AsyncTasks into a single one (improves code quality)
- [x] **EditNoteActivity:** ensures, that auto-save is not called, if the previous auto-save is still running. In addition, start next auto-save not before a small delay has passed. (This could fix #76)

#### Other Issues I've Found (not covered by this pull request)
- There is no conflict resolution for the case that a Note is locally and remotely edited in parallel (e.g. during a longer time without network connection). Currently, local changes overwrite remote changed. We have to discuss what would be the desired behavior. See also #66.
- The `ItemAdapter` should use a `CursorAdapter` which is updated automatically if the backend data is changed.
- In order to save traffic during synchronization, `pullRemoteChanges()` could use the `exclude` parameter of the [Notes API](https://github.com/owncloud/notes/wiki/Notes-0.2) in a first step to check which notes changed without downloading unchanged notes. Only in the second step, the whole note content would be downloaded from the changed note. This could save traffic, however, it has to be investigated if the effort worth it (what is the typical/topmost number and size of notes?).


stefan-niedermann:
I think i have some free time at this weekend. I'll test this, merge this and if all is going well, i'll build a new version for F-Droid.

stefan-niedermann:
looks good. shall i wait for the three open tasks or do prefer to merge now?

korelstar:
I tested my branch again and I found an embarrassing bug. Unfortunately, I can't fix before Monday. So please wait a few days. I'll write a comment here when I'm finished.

korelstar:
ready to merge :-)

moshpirit:
\o/ wooohoo! Thanks a lot for the support!!

stefan-niedermann:
@korelstar just let me say a huge **thank you** again - i just had time again to watch your code and it looks awesome.
I want to tag a new version for F-Droid since the current version 0.8.1 is buggy. One little thing i noticed: If you are doing a clean install of the app (or purge any data of it) and start it, you will get into the settings activity - that's fine, but it trys to sync and fails (and shows an error message).
I don't have a problem if it trys a sync on first start-up, but it should at least not display an error message.

korelstar:
I will have a look at this issue and fix it as soon as possible, but I can't tell when, since I just changed my residence and I haven't access to my working station yet.

PR #138, opened 2016-10-05
Extend "excerpt" with text from the following lines

korelstar:
Currently, only the second line is shown as `excerpt` in the list of notes. However, often the second line is not that interesting. So I propose to show more text from the note.


stefan-niedermann:
You are limiting to 200 characters, am i right?

korelstar:
Yes, that was arbitrary and we can change this. I just wanted to bound the length, because notes can be several KBs long and there is no need, that the excerpt has this length, too. Hence, it's just a performance issue.

PR #139, opened 2016-10-05
Load version name dynamically from package

korelstar:
I think with this, it will be much easier to release a new version and maintaining translations will result in less conflicts.


stefan-niedermann:
Agree! Thanks

PR #142, opened 2016-10-17
Persistant notification for creating new notes

kas70:
Can be enabled with a checkbox preference.

I did this for myself because i needed a quick way for creating new notes. Maybe this is useful for you.
@stefan-niedermann

### DIFF
--- a/app/src/main/java/it/niedermann/owncloud/notes/android/activity/NotesListViewActivity.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/android/activity/NotesListViewActivity.java
@@ -1,5 +1,7 @@
 package it.niedermann.owncloud.notes.android.activity;
 
+import android.app.NotificationManager;
+import android.app.PendingIntent;
 import android.app.SearchManager;
 import android.content.Intent;
 import android.content.SharedPreferences;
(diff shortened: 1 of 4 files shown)


stefan-niedermann:
Thank you for contribution @kas70 !
@jancborchardt your opinion please: Do you think this is a setting, that will be used wildly? I'm struggling with myself because even it is a great feature i am not sure if it should be built in.

jancborchardt:
Hm, usually this would be solved with a widget, no? Isn’t hacking the notifications for that a bit non-standard?

I would definitely have it as a setting only instead of the default.

kas70:
Persistent notifications are often used for providing access to quick actions. Further, i'm personally no friend of widgets, but that's a matter of taste.

jancborchardt:
Ok, seems fine then as an off-by-default setting. :)

stefan-niedermann:
Okay, i think i will merge it even if i feel a bit uncomfortable with it.
Don't get me wrong, i just don't want to create too much settings for user which only maybe 5 % will use. (Maybe we can think about adding some Tabs to the Settings Activity, or a "advanced options"-drop down so most user will only see the url & credentials stuff - if we get more options).

jancborchardt:
@stefan-niedermann I totally agree, I’m also for less settings and stuff to maintain.

PR #149, opened 2016-10-27
Consolidate repeated and untranslatable names

korelstar:
I'm not sure if this is too cheeky, but after looking at the [contributors statistic](https://github.com/stefan-niedermann/OwnCloud-Notes/graphs/contributors), I've added myself to the developers. :sunglasses:

In order to be also productive, I've removed all (in my point of view) untranslatable (and hence identical for all languages) personal names from the translations and set `translatable="false"` for them. This should increase maintainability.


stefan-niedermann:
Sounds great! I thought about displaying in every language only the responsible translators to reduce the maintenance effort...

korelstar:
> I thought about displaying in every language only the responsible translators to reduce the maintenance effort...

Yes, why not. Sound reasonable.

Issue #118, opened 2016-06-19
Every touch brings to the top


sprnza:
Hi there!
thank you for the great app!
I've faced with a very annoying bug. Every tap on a note preview brings me to the top of it. The only chance to have it scrolled to the bottom is to do it continiously swapping down and catching while it moves.
I use 0.8.1 from f-droid

stefan-niedermann:
Hm, i'm sorry i can not reproduce it. Can you reproduce with the current master branch? Otherwise you could wait for 0.9.0 which has many overall code improvements

moshpirit:
Came here because this is happening to me on last version on F-droid: 0.9.0.

Also, this bug deselect all notes selected. So if you select like your last 15 notes, it brings you to the top (I think it's an automatic sync) and deselect (even when the top bar shows "15 selected")

Android 6.0.1

stefan-niedermann:
Tribute to @korelstar, should be fixed in v0.10.0 (will probably released after the next server-side-app release + some time for testing and bugfixing).

Issue #92, opened 2016-02-17
Option to change sort method


dasmaetthes:
Would be nice to sort notes by date of last edit or alphabetically

fxedel:
:+1:

stefan-niedermann:
@jancborchardt your opinion pls - an option in the settings, default "last modified"?

fxedel:
My summary of possible sorting options:

- Last edit
- Created
- Alphabetically (name)
- Manually (needs changes in how the notes are saved)

jancborchardt:
Here’s my opinion on sorting of notes:

- Last edit: How we currently do it. It continuously changes sort order, but keeps the relevant stuff at the top.
- Created: I really don’t see how sorting by the date of _creation_ is relevant. A few months down the line, the sorting will make no sense in day-to-day working.
- Alphabetically (name): Alphabetical sorting in general is the lowest common denominator when it comes to sorting. I don’t think it’s a good idea to sort like this in general, especially not in a Notes app where titles can be anything. Look at Google Keep, Evernote, etc – they all do sort by modified date.
- Manually (needs changes in how the notes are saved): I would really not do this. It requires manual work on every single user’s side, and it’s likely that people will only »find« it by accident. Also, what’s the usecase? Stuff you would like to see sorted up top is likely already up top with the sorting by modified date.

I’d suggest that we keep it very simple and only have the sorting by modified date, like now.
I can maybe, very maybe, understand the need for alphabetical sorting, but I’d wait with that until a lot of people have very good reasons we should build this in.
Cause right now sorting by modified date seems clearly superior, especially on the go. With alphabetical sorting, you basically have to scroll for mostly every note. With modified date, the most important stuff is on top automatically. Keep in mind this is a mobile app used on phones etc.

@dasmaetthes @fxedel what’s your use-case and why do you need a different sorting method?

And btw @stefan-niedermann a slightly related idea: Is there an API to check if a note was favorited (has a star) on the server? We could use that info and always sort the starred files up top, also adding a small indicator.

jancborchardt:
And about ascending/descending: Sorry but that makes absolutely no sense for the first 3 methods. For modified & created date this would show a stale list of the latest notes which is not useful, and for alphabetical sorting it would show things in a non-standard order with the ZYX stuff first.

We should always think thoroughly about anything we add to the app. Especially because it’s on mobile, and especially also about any setting and if it’s needed.

X4fyr:
[...]

PR #151, opened 2016-11-06
New feature: set a note as favorite (star/unstar)

korelstar:
Setting a note as favorite was already introduced on the server side (owncloud/notes#248 and nextcloud/notes#4) and can be used to stick important notes to the top of the list in order to find them easier (cp. #92). My changes here introduce this feature for the android app.

Beside this, I did some necessary refactoring in the `NotesListViewActivity`/`ItemAdapter`: (re-)loading the list of notes is now done in an `AsyncTask` and reuses the `ItemAdapter` instance (instead of creating a new instance on every refresh). As a result, the list of notes does not jump to the top anymore (cp. #118).

The implementation of the favorite feature is straight forward and has no special treatment for old(*) server app versions (i.e., versions that do not support the favorite feature). As a result, if a user sets a note as favorite, the star disappears after the next synchronization with an old server app. This behavior can be improved with some effort:

a) activate the feature only on demand => add a checkbox to the settings activity and only show the favorite feature, if this checkbox is activated

b) detect server features automatically => like a), but instead of setting the checkbox by the user, the app has to check if the server uses an appropriate version. This requires changes in the server API (e.g. an `info` URL which returns a JSON with the version number) and its unclear, at which point of time the app should check this in order to detect updates of the server app.

c) activate the feature, but keep the local state if the server result does not contain any information about favorites => requires some changes in the app; problem: when the user updates the server app he/she will lose the local favorites setting (the server favorites will overwrite the local favorites). In order to improve this, some more effort is needed, e.g. a new DB column is required, which notices if the local meta data (i.e. favorite setting) was changed and not yet synchronized. Therefore, this solution requires the most effort (I didn't see this in my first estimation: https://github.com/stefan-niedermann/OwnCloud-Notes/issues/92#issuecomment-256010808).

As an overall consequence, I would leave this as is -- for now. If a better downward compatibility is required, then we can do this in a separate commit/pull request.

(*) today, "old server app" means also the current official version of the server app. For now, the favorite feature is only in the git `master` branch and is waiting for the next release.


stefan-niedermann:
**TL;DR**: Merged!

Great work @korelstar as usual and thanks for the detailed documentation.

I agree with you in all points: **c) is the preferred option**. While no one has time to implement this, the **current state is okay**. I assume i will release v0.10.0 after the next server-app-version has released (+ some time for bugfixing and testing).

Assuming everyone keeps his instance up2date, we won't have problems with down-ward-compatibility.
What we could do at this point:

@Henni Can you tell us, how you are dealing with down-ward-compatibility? Which versions do you support? Only the current?

PR #153, opened 2016-12-09
fix onActivityResult throws exception if activity is re-created

korelstar:
see https://github.com/stefan-niedermann/OwnCloud-Notes/issues/2#issuecomment-265564930 and https://github.com/stefan-niedermann/OwnCloud-Notes/issues/2#issuecomment-266115838. Only update the changed note if the position exists in the list. If this is not the case, this is due to recreation of the activity and the list will be updated by the background task (with the changed values).


stefan-niedermann:
Merged into `RxMarkdown`-Branch in eeec53a66dbc30413be4c333a5a2777ca0c546f9 and 2db5e81844361188784c0b6a43f1cccd5959e25e

Issue #147, opened 2016-10-26
Move from ownCloud to Nextcloud


stefan-niedermann:
The server side app [moved to nextcloud](https://github.com/owncloud/notes/issues/235).
- [x] Adjust theme to match the lighter blue color set of nextcloud
- [x] Adjust icon
- [x] Refactore code from owncloud to nextcloud
- [x] Adjust README.md in this repository

And if we want to rename the repository itself, too:
- [x] Clone repository to /nextcloud-notes
- [x] Port all issues to the new repository
- [ ] Adjust links on Google Play Store and F-Droid

korelstar:
I think, that's a political decision. For now, the app works with both "clouds" and we don't know what will happen with the "old" owncloud app (maybe all changes in the nextcloud notes app will be backported to owncloud, then this app will keep being compatible).

But if you want to rename this app and move it's repository, please note that you can rename the repository and github will make respective redirects for the git url, issues, etc.: https://help.github.com/articles/renaming-a-repository/

stefan-niedermann:
Links are already adjusted and the only remaining task is new artwork, colors, icons and text in the Play Store - but this has nothing to do with this project in general.

PR #159, opened 2016-12-21
Refactoring: OwnCloudNote -> CloudNote

korelstar:
see #147

This only changes the name of the class `OwnCloudNote`. The package is still `it.niedermann.owncloud.notes`, since changing this will have more impacts.


stefan-niedermann:
Package name will remain `it.niedermann.owncloud.notes`.

PR #172, opened 2017-01-23
trim leading and trailing spaces in the URL

AndyScherzinger:
I added trim() to the URL handling since user typos like a space can be seen (in the text field) but will block the server check. So this change is just for UX reasons. 😸 

**Android version:** Android 7.1.1

**Device**: Nexus5X

**System language**: language English, Location Germany (to have English language and the Euro)

**App version:** git master 23/01/2017

**App source:** git master

**Steps to reproduce:**
  1. open the app
  2. try a login and add a trailing space to the URL


stefan-niedermann:
Thanks :)

AndyScherzinger:
Happy to contribute :)

PR #178, opened 2017-02-21
Fix string-resources / Adjust lint-options to make a standard gradle-build possible

4ch1m:
* double-quotes don't need to be escaped within the xml-resource-files (currently the build will fail because the xml-resources can't be parsed properly due to these escapings)

* there are several language-resource-files/translations with missing entries; lint will complain about this and cancel the build

PR #180, opened 2017-02-26
Fix and clear up translations

### LINKED ISSUE #175: Some issues on translations/Transifex
- [x] In some languages (e.g. `es`), quotes in hyperlinks are escapes with backslashes which breaks the build: e.g., https://github.com/stefan-niedermann/nextcloud-notes/blob/3647ec3c3524ce409bc51e61113c53b215e38d5b/app/src/main/res/values-es/strings.xml#L77 -- However, I don't know where this comes from. Maybe we have to use `&lt;` instead of `<` for HTML tags inside of a string, but that's only a hypothesis.
- [x] Constant parts of strings (e.g. URLs) should be moved into separate (untranslatable) string. This will simplify changes in these parts (i.e. in `about_source`, `about_issues`, `about_app_icon_disclaimer`, `about_icons_disclaimer`, maybe also the license name in `about_app_license`).
- [x] Some of these URLs point to an outdated location, so a change is definitely required (`about_source`, `about_issues`).
- [x] I propose to merge the two sections about icon into one (`about_app_icon_disclaimer`, `about_icons_disclaimer`). Strings can be concatenated.
- [x] The list of translators (`about_translators`) is obsolete. I propose to change the string to `Nextcloud community on Transifex`.
- [x] Plurals are currently in a separate file (`plurals.xml`). However, there is no need for this and it causes that there are two separate resources in Transifex, which bloats the project. I propose to integrate plurals into `strings.xml`.
- ~~There is an XML schema in order to provide more information about untranslatable parts of a string (especially placeholders) which can be helpful for translators (see [docs](https://developer.android.com/distribute/tools/localization-checklist.html)), e.g. `<xliff:g id="app_version" example="v2.1.0">%1$s</xliff:g>`. We could use it, but I don't know how it works with Transifex!~~ *doesn't work with Transifex export*
- [x] (by @stefan-niedermann ) `Contribute`-tab should get a new section "Translate"

Fixed in PR #180.

korelstar:
This will fix #175. See there for more details.

Since this have to be commited before @nextcloud-bot exports new translations from transifex in the next night, I will merge this in this evening at the latest. If someone wants to review this, please be quick :wink: 

One note: Now, the `<` are escaped by `&lt;` in the string resources. Therefore, we have to use `Html.fromHtml(String)` which is replaced by `Html.fromHtml(String, int)` in API 24. For better readability, I have introduced a helper class which takes care of this distinction.

### DIFF
--- a/app/src/main/java/it/niedermann/owncloud/notes/android/fragment/about/AboutFragmentContributingTab.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/android/fragment/about/AboutFragmentContributingTab.java
@@ -2,20 +2,21 @@ package it.niedermann.owncloud.notes.android.fragment.about;
 
 import android.os.Bundle;
 import android.support.v4.app.Fragment;
-import android.text.method.LinkMovementMethod;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.TextView;
 
 import it.niedermann.owncloud.notes.R;
+import it.niedermann.owncloud.notes.util.SupportUtil;
 
 public class AboutFragmentContributingTab extends Fragment {
     @Override
     public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState) {
         View v = inflater.inflate(R.layout.fragment_about_contribution_tab, container, false);
-        ((TextView) v.findViewById(R.id.about_source)).setMovementMethod(LinkMovementMethod.getInstance());
-        ((TextView) v.findViewById(R.id.about_issues)).setMovementMethod(LinkMovementMethod.getInstance());
+        SupportUtil.setHtml((TextView) v.findViewById(R.id.about_source), R.string.about_source, getString(R.string.url_source));
+        SupportUtil.setHtml((TextView) v.findViewById(R.id.about_issues), R.string.about_issues, getString(R.string.url_issues));
+        SupportUtil.setHtml((TextView) v.findViewById(R.id.about_translate), R.string.about_translate, getString(R.string.url_translations));
         return v;
     }
 }
--- a/app/src/main/java/it/niedermann/owncloud/notes/android/fragment/about/AboutFragmentCreditsTab.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/android/fragment/about/AboutFragmentCreditsTab.java
@@ -3,26 +3,27 @@ package it.niedermann.owncloud.notes.android.fragment.about;
 import android.content.pm.PackageManager;
 import android.os.Bundle;
 import android.support.v4.app.Fragment;
-import android.text.method.LinkMovementMethod;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.TextView;
 
 import it.niedermann.owncloud.notes.R;
+import it.niedermann.owncloud.notes.util.SupportUtil;
 
 public class AboutFragmentCreditsTab extends Fragment {
     @Override
     public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState) {
         View v = inflater.inflate(R.layout.fragment_about_credits_tab, container, false);
-        ((TextView) v.findViewById(R.id.about_maintainer)).setMovementMethod(LinkMovementMethod.getInstance());
         String versionName;
         try {
             versionName = "v"+getActivity().getPackageManager().getPackageInfo(getActivity().getPackageName(), 0).versionName;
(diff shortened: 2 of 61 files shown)


korelstar:
@MorrisJobke
I did some structural changes on our translations (e.g. change `<` to `&lt;`, see #175) and changed the source (`values/strings.xml`) as well as the translations (`values-*/strings.xml`). @nextcloud-bot uploaded the source to Transifex but did not uploaded the new translations. Is this intended behavior?

If yes, is there a way for us to upload my automatically changed translations to Transifex? Because otherwise, the translators have to retranslate many strings they already translated before.

MorrisJobke:
> If yes, is there a way for us to upload my automatically changed translations to Transifex?

No - there is no way to sync the translations, because this would mean really complex collision detection. So the translators need to retranslate. Anyways transifex is smart enough and shows the old version ;)

korelstar:
@MorrisJobke 
Damn, then much of my work was useless. Btw., is @nextcloud-bot open source? That would be helpful to understand the exact way of working.

And another thing: @nextcloud-bot broke our build due to incompatible escaping (see #175 and #178). We have fixed this in changing the source language file, but breaking the build should nevertheless prohibited (see https://github.com/stefan-niedermann/nextcloud-notes/pull/178#issuecomment-282106283). It would be cool if @nextcloud-bot would at least do some kind of syntax checking before committing to `master` branch, or better: only commit if the build has passed successfully.

MorrisJobke:
> Damn, then much of my work was useless. Btw., is @nextcloud-bot open source? That would be helpful to understand the exact way of working.

Yes it is:

This is the script: https://github.com/nextcloud/docker-ci/blob/master/translations/handleExternalAndroidTranslations.sh (called with `stefan-niedermann nextcloud-notes` as argument list)

This is run in a docker container (which contains all the tools): https://github.com/nextcloud/docker-ci/blob/master/translations/Dockerfile

So we basically only use the tx tools to do so. 




> And another thing: @nextcloud-bot broke our build due to incompatible escaping (see #175 and #178). We have fixed this in changing the source language file, but breaking the build should nevertheless prohibited (see #178 (comment)). It would be cool if @nextcloud-bot would at least do some kind of syntax checking before committing to master branch, or better: only commit if the build has passed successfully.

Feel free to implement this, but it needs to do a branch, push the branch, create a pull request, wait for the pull request to finish successfully and merge. And this everyday. It's on one hand a lot of scripting around the pull request handling and on the other side a lot of open pull requests (one per day).

We also never noticed any problems beside on android regarding broken syntax. Is there maybe an easy linter we can execute right after pulling the translations? On the other side somebody then needs to get notified about this somehow to fix it.

korelstar:
Thanks for the links, @MorrisJobke 


>Feel free to implement this, but it needs to do a branch, push the
>branch, create a pull request, wait for the pull request to finish
>successfully and merge. And this everyday. It's on one hand a lot of
>scripting around the pull request handling and on the other side a lot
>of open pull requests (one per day).

I think one could go without the pull request and only push to a separate branch. The script just has to do a hard reset of that branch to master before committing. Then one could easily merge the translations branch manually on demand (e.g. before a new release). This would also reduce the number of commits by @nextcloud-bot (reduces Git history pollution).

The commands in the script would be:
```
git checkout $BRANCH
git reset --hard master
tx push -s
tx pull ...
git add .
git commit -a ...
git push -f
```
(We could do this only if the script is called with a third argument, so every repository can decide individually which approach should be used.)

What do you think, @MorrisJobke and @stefan-niedermann? (If you give positive feedback, I'll make a pull request to change the script.)


> Is there maybe an easy linter we can execute right after
>pulling the translations? On the other side somebody then needs to get
>notified about this somehow to fix it.

Yes, Android has a linter. But I think the best would be to use the existing CI configuration of the respective repository.

MorrisJobke:
@korelstar This doesn't work out for out many repos, so we will do it always directly on master. Feel free to create a copy of the script that does this (instead of the argument).

Then I can run this for this specific repo here on the other script.

@stefan-niedermann @korelstar Currently the transifex push fails, because the nextcloud-bot get rejected with "protected branch hook declined".

korelstar:
> @korelstar This doesn't work out for out many repos

Why?


> @stefan-niedermann @korelstar Currently the transifex push fails,
> because the nextcloud-bot get rejected with "protected branch hook
> declined".

Hmm. Did you change something on the access rights, maybe in conjunction 
with Travis, @stefan-niedermann?

MorrisJobke:
> Why?

Because all we would do is to merge translation branches. We have better stuff to do. In 99.9% of all cases pushing to master works well. And if something goes wrong we need to investigate anyway, why it failed.

PR #182, opened 2017-02-28
update RxMarkdown to 0.0.8-beta

korelstar:
looks definitely nicer! :smile:

However, the bug yydcdut/RxMarkdown#19 still occurs. :cry:


stefan-niedermann:
Well, i am very happy that the author of the lib responded and it seems like something is happening. Maybe we just have to bit a little more patient :)

PR #188, opened 2017-03-04
[Work in Progress] Add travis support and adjust lint settings

stefan-niedermann:
cc @4ch1m and @korelstar this branch is for integrating Travis. I have cleaned up the history in master and i continue with my experiments over here :) Any input is welcome as always.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,12 @@
+language: android
+
+jdk: oraclejdk8
+
+android:
+  components:
+    - tools
+    - build-tools-25.0.1
+    - android-25
+    - extra-android-m2repository
+
+script: ./gradlew testDebug
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -17,6 +17,10 @@ android {
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
         }
     }
+    lintOptions {
+        disable 'MissingTranslation'
+        abortOnError false
+    }
 }
 
 dependencies {


korelstar:
Sorry, I have currently no time to look into the details. But it looks like you solved the license issue. 👍 
So, only the emulator doesn't work. But what do want to do with the emulator? Do we really need it? The most important thing is to check if the build is successful. Later, we can add some unit tests. But as far as I know, all this can be done without the emulator (see e.g. [K9mail](https://github.com/k9mail/k-9/blob/master/.travis.yml)).

stefan-niedermann:
Right, i solved the license issues and the only thing that does not work yet is the emulator.
And i agree that this is currently not necessary. I am wondering how k-9 builds work without license-issues since they have no `license`-keyword in their yml-file. Hm. Still experimenting. Will try to use exact same file as k-9 and see what happens :)

korelstar:
I did some experiments, too (on the branch `travis-test`). I got a successful build with the following `.travis.yml`: https://github.com/stefan-niedermann/nextcloud-notes/blob/8a9a397520e2e4abd746af5f60e1676b5828c56a/.travis.yml, see https://travis-ci.org/stefan-niedermann/nextcloud-notes/builds/209390323

But: The existing unit tests were not started, so the build result is always successful, even if the tests fail. But that would be fine for the first step ... We can work on that later ...

korelstar:
worked! :grin:

stefan-niedermann:
Well then - looks good :) Wanna merge it yourself into master?

PR #193, opened 2017-03-19
Accept user-installed certificates with Nougat

ilpianista:
Please merge this as with Android 7 you need to explicitly allow user installed certificates (see https://developer.android.com/training/articles/security-config.html).


stefan-niedermann:
Thank you for contribution :). I will read the linked document and then merge if everything is okay.

PR #195, opened 2017-03-25
Update strings.xml

rakekniven:
Removed brand owncloud. Could be nextcloud as well. Suggest to remove cloud name.


stefan-niedermann:
Thank you for contribution. Aggree to remove cloud name where possible.

Issue #197, opened 2017-03-27
Shortcuts for Android 7


stefan-niedermann:
https://developer.android.com/guide/topics/ui/shortcuts.html#static

![device-2017-04-09-124450](https://cloud.githubusercontent.com/assets/4741199/24836692/9a439258-1d22-11e7-93ac-3a5016a68769.png)

Pull Request: https://github.com/stefan-niedermann/nextcloud-notes/pull/201

PR #201, opened 2017-03-31
#197 Shortcuts for Android 7

stefan-niedermann:
Added icon and shortcut.

~~To do: On finish creating a new note, one should go back to list view instead of closing the app completely.~~ This is no urgent topic and can be solved later.


korelstar:
- Works in my emulator.
- Do you really need `shortcutDisabledMessage` in `shortcuts.xml` (and `shortcut_create_disabled` in `strings.xml`)? There is no need to disable the shortcut and it seems, that the attribute is optional (but I'm not sure).
- The icon doesn't look nice. At least, it should be colored according to the theme.

stefan-niedermann:
Agree on both points and fixed.

device-2017-04-09-124450

PR #202, opened 2017-04-02
handle links between notes inside app

fabienli:


stefan-niedermann:
Can you explain, what you are implementing here? it looks like a custom handler foe note:// protocol, so one can provide links at websites that open the notes app when you click on it?

fabienli:
On desktop, I'm using QOwnNotes (from www.qownnotes.org) which can create links between notes.
On nextcloud-notes, the link is then appearing but by default, nothing happens when clicking. I've tried then to make the link active. This Pull Request currently fits my personal needs, let me know if this can be useful for the whole project.

stefan-niedermann:
I am still trying to get it. Please tell me, if i understood correctly:

You have got two notes, `Note A` and `Note B`.
Content of `Note A` is

```
This is a link to [Note B](2435)
```

Clicking on the created link opens `Note B`.

Correct so far?

Then i have two questions
1. What has to be the "link"? The id of the note? Title? Filename?
2. I like this feature. is it implemented in the server side app?

stefan-niedermann:
ping @fabienli

fabienli:
Correct. Just the naming convention of the link needs to be precised:

1. The link is the filename (spaces replaced by underscores). 
If you have two notes "note A" and "note B", you can put in the first one: 
[Link to B](note://note_B)
And this would enable to jump into the note B
2. I don't think anything is implemented in server side, I'm using this based on qOwnNotes as Desktop app. I'm not using the notes on the webserver actually.

stefan-niedermann:
cc @Henni would you accept this implementation on the server side? If yes: we should define a and commit to a convention as proposed by @fabienli (though i'd rather see a `-` than an `_`.
Another option would be to add it as a seperate field to the api, so not every client has to implement its own code?

Henni:
Seems like a good idea.
But I'm not sure what the link should be. The filename changes to often and isn't immediately visible in the app.
What do you mean by 'a separate field' in the API?

stefan-niedermann:
@Henni i still think the idea is quiet nice. Every note has an id, why not use it?

stefan-niedermann:
Closing for now due to inactivity.

Please reopen if needed in the server repo at https://github.com/nextcloud/notes and if it is implemented there, we can handle it on the android client as well.

frankgerhardt:
What a pity that this PR has been closed without being merged.

stefan-niedermann:
So let's discuss this topic again - it's a premise for me that the server app will implement this feature too. And i am pretty sure we can reuse the code from this PR.

frankgerhardt:
Ok, I think the main distinction is between outbound links like http:, mailto: etc. and internal links where the target is another note. In the files app and the markdown editor the file has an extension, .md probably. In the notes app the notes have just a name without an extension. But the names can have spaces. I'd like to write internal links like this:

```
[link to other note](Other Note)
```

As far as I can see it two things need to be done, 1. the editor has to render the link that the browser understands, maybe Control-click, and links to a valid URL that opens the note in the note editor, and 2. the server needs to serve these URLs. 

I'm not experienced with PHP and I'm not sure how much I can help. I have the plan though to write an app for NC some day. Because of that I'm interested. At least I could set up a test environment and help with testing.

stefan-niedermann:
The problem is, the file name equals the first line of the note and can change easily over the time.

Therefore, it might be a way more stable way to link to the `noteId` which doesn't change over the time.

frankgerhardt:
As an author of a note I would not know the id and looking it up (or auto-completion? ;-) ) would make the authoring experience painful. It would interrupt the writer's thought process. With auto-completions it would not be an issue.

stefan-niedermann:
The android app already supports context based formatting (when you select some text, you can already insert the markdown for a link). this can be extendet to pick a note.

The webinterface is also working on some kind of context based formatting which would allow them to do some similar picker.

For advanced users in the webinterface who want to directly write the markdown: the `noteId` is part of the url.

stefan-niedermann:
Just for documentation: Simple links between notes have been implemented recently. They base on the server-id of the note, so just open a note in the browser, get the last part of the url (only the number) and link it, like: `[My fancy note](12345)`.

NESC1US:
thanks for your work, i see the point of having ids as titles change, but I would still prefer having the address based on title (filename) which would then allow notes to be exportable and importable between nextcloud instances (and majority of other markdown applications) while keeping the links between them.

stefan-niedermann:
@NESC1US in this case you will need a fully qualifies URL containing the other instance. You can find  fully qualified URL in the files app when sharing a note file publicly.

NESC1US:
I dont think that I explained it well.. i meant to argue for wiki-links-like functionality with relative links refering to a file name - something like github and gitlab have:
- github uses [relative paths in readme.md](https://github.blog/2013-01-31-relative-links-in-markup-files/)
- gitlab supports same [here](https://about.gitlab.com/handbook/markdown-guide/#relative-links).
when those files are synchronized via webdav then link keep working without nextcloud - this allows it to be used in many markdown capable applications ([ghostwriter](http://wereturtle.github.io/ghostwriter/), [ReText](https://github.com/retext-project/retext), [ZIM desktop wiki](https://zim-wiki.org/index.html) and many more)

the issue with changed titles seems solvable by keeping references to links and update them when the title changes.

stefan-niedermann:
Thank you for the explenation. However i think this should be discussed in the [corresponding issue of the server app](https://github.com/nextcloud/notes/issues/616) because it should be the same behavior on all platforms and the server app is leading.

We can (and should) discuss this again here, if and as soon this has been implemented in the server app.

Issue #199, opened 2017-03-31
App crashes when accessing note and go back to list


stefan-niedermann:
It is very hard to reproduce this issue. Appears with current master and v0.10.0.

Steps:

1. Open any note from the list view
2. Close this note as quick as possible (using native back-button or top left arrow button does not matter)
3. Repeat Steps 1 - 2 until the app crashes (10 - 15 times at my device)

CC @korelstar 

```
03-31 08:55:11.909 2691-2725/? D/SyncTask:    Process Remote Note: R20  (*)  / MarkDown Test (2017-03-22 20:08:07 / null)
03-31 08:55:11.910 2691-2725/? D/SyncTask:    ... found -> Update
                                           
                                           --------- beginning of crash
03-31 08:55:11.915 2691-2725/? E/AndroidRuntime: FATAL EXCEPTION: AsyncTask #3
                                                 Process: it.niedermann.owncloud.notes, PID: 2691
                                                 java.lang.RuntimeException: An error occurred while executing doInBackground()
                                                     at android.os.AsyncTask$3.done(AsyncTask.java:325)
                                                     at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:354)
                                                     at java.util.concurrent.FutureTask.setException(FutureTask.java:223)
                                                     at java.util.concurrent.FutureTask.run(FutureTask.java:242)
                                                     at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:243)
                                                     at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1133)
                                                     at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:607)
                                                     at java.lang.Thread.run(Thread.java:761)
                                                  Caused by: java.lang.IllegalStateException: attempt to re-open an already-closed object: SQLiteDatabase: /data/user/0/it.niedermann.owncloud.notes/databases/OWNCLOUD_NOTES
                                                     at android.database.sqlite.SQLiteClosable.acquireReference(SQLiteClosable.java:55)
                                                     at android.database.sqlite.SQLiteDatabase.updateWithOnConflict(SQLiteDatabase.java:1548)
                                                     at android.database.sqlite.SQLiteDatabase.update(SQLiteDatabase.java:1525)
[...]

PR #203, opened 2017-04-02
trying to solve issue #199

fabienli:


stefan-niedermann:
Great that you are able to contribute here! Keep us up to date so we may can help and or review your fix. :)

cc @korelstar

stefan-niedermann:
@fabienli ready to test?

stefan-niedermann:
@korelstar so you are suggesting that we remove all `db.close()` calls?
You explained the effect is the same, i see. I am fine with both variants. If you want to create a separate pull request with the proposed changes i will merge it, otherwise i will merge this PR. Anyway, this issue is annoying and i want to get rid of it.

korelstar:
Yes, I suggest to remove that. I will create a PR soon.

stefan-niedermann:
Fixed with #212. Thank you very much @fabienli for bringing us on the track :) 👍

PR #207, opened 2017-04-18
vibrate on list item movement

korelstar:
In order to decrease the likelihood to delete a note accidentally, a short vibration warns the user when she/he begins to move a note in the note list.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -5,6 +5,7 @@
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.USE_CREDENTIALS" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
+    <uses-permission android:name="android.permission.VIBRATE" />
 
     <application
         android:allowBackup="true"
--- a/app/src/main/java/it/niedermann/owncloud/notes/android/activity/NotesListViewActivity.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/android/activity/NotesListViewActivity.java
@@ -3,11 +3,13 @@ package it.niedermann.owncloud.notes.android.activity;
 import android.app.NotificationManager;
 import android.app.PendingIntent;
 import android.app.SearchManager;
+import android.content.Context;
 import android.content.Intent;
 import android.content.SharedPreferences;
 import android.graphics.Canvas;
 import android.os.AsyncTask;
 import android.os.Bundle;
+import android.os.Vibrator;
 import android.preference.PreferenceManager;
 import android.support.design.widget.Snackbar;
 import android.support.v4.view.MenuItemCompat;
@@ -56,6 +58,7 @@ public class NotesListViewActivity extends AppCompatActivity implements
     private ActionMode mActionMode;
     private SwipeRefreshLayout swipeRefreshLayout = null;
     private NoteSQLiteOpenHelper db = null;
+    private Vibrator vibrator = null;
     private SearchView searchView = null;
     private ICallback syncCallBack = new ICallback() {
         @Override
@@ -81,6 +84,8 @@ public class NotesListViewActivity extends AppCompatActivity implements
         }
 
         setContentView(R.layout.activity_notes_list_view);
+
+        vibrator = (Vibrator) getSystemService(Context.VIBRATOR_SERVICE);
 
         // Display Data
         db = NoteSQLiteOpenHelper.getInstance(this);
@@ -275,6 +280,9 @@ public class NotesListViewActivity extends AppCompatActivity implements
         listView.setAdapter(adapter);
         listView.setLayoutManager(new LinearLayoutManager(this));
         ItemTouchHelper touchHelper = new ItemTouchHelper(new SimpleCallback(0, ItemTouchHelper.LEFT | ItemTouchHelper.RIGHT) {
+
+            private boolean active=false;
+
             @Override
             public boolean onMove(RecyclerView recyclerView, RecyclerView.ViewHolder viewHolder, RecyclerView.ViewHolder target) {
                 return false;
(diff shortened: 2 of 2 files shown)


stefan-niedermann:
Hm.. this sounds like the vibration could be caused very often. Wouldn't it be better to vibrate when the deletion is done? so basically change from "Hey, you are swiping" to "Hey, if you want to restore it - here is the toast! :-)"

korelstar:
My intention was to inform the user that there is a special action, *before* the action was executed.

The app OpenTasks has a similar vibration, but there, it is triggered already when doing a selection. That is annoying since it happen much more often. In my approach, the vibration is only triggered on movement. I think this is a good approach.

Please try it out by yourself and check how often the vibration is triggered when you didn't want to delete a note.

stefan-niedermann:
cc @jancborchardt can you please add your opinion? vibrating on a danger action seems valid for me. but the question is: vibrate *before* the user has done it or *after*?

jancborchardt:
Hm, vibration in apps for actions like these is very unusual. Can you give any examples of _widely-used_ apps doing this?

korelstar:
"Widely-used" is hard to say, but I've seen this in OpenTasks. But IMHO more important than "widely-used" is if it's useful! And I would say: yes.

jancborchardt:
Well this is a strange pattern though. Vibration is used for notifications. For example I have vibration for typing disabled and it would be strange to have it in this one app.

A better approach would be to simply offer undo on note deletion. :)

korelstar:
Just read this: https://docs.nextcloud.com/server/12/developer_manual/general/codingguidelines.html
> Do not adapt broken concepts (for example design of desktop apps) just for the sake of consistency. We aim to provide a better interface, so let’s find out how to do that!

:wink:

jancborchardt:
@korelstar I wrote those. ;) Vibration is not a desktop app pattern.

Just let me know some other popular apps which use the same pattern so we can be sure people are reasonably used to it. Otherwise I would say this is pretty strange.

Since the goal is to minimize accidental deletion, I would rather say there should be something like undo deletion as implemented by Google Inbox or Android or iOS. There are just a few seconds (5–7) in which the action is not carried out yet and while you can undo it. For that I recommend reading »[Never use a warning when you mean undo](https://alistapart.com/article/neveruseawarning)« by Aza Raskin, who was heading Firefox UX at the time.

korelstar:
Guessed that you wrote this :stuck_out_tongue_winking_eye: 

However, there is already an undo function in this app. And I'm reasonably introducing a vibration when moving the list item. This is like it's vibrating when you move a heavy stone from another stone. It gives you a feeling that this will have an effect. It's just the small thing, that makes life better, without being annoying for people who do this routinely. Nothing becomes worse, just an upgrade. :+1: 

One reference is [OpenTasks](https://play.google.com/store/apps/details?id=org.dmfs.tasks).

jancborchardt:
@stefan-niedermann what do you think? And our Android experts @AndyScherzinger @tobiasKaminsky @mario?

stefan-niedermann:
to be honest, I still think it would be a better approach if we could desensitize the swipe gesture. I don't like the vibrate action though i agree that the swipe-delete happens to easy.

tobiasKaminsky:
I am using the app since many months and never realised that there is "swipe right" to delete, so you can guess that it did not happen accidentally for me ;-)
I also think that the better approach is to desenitize the swipe to delete and not to add a less known UX feedback.

AndyScherzinger:
Looking at Gmail there is also the swipe to delete offering an undo action which seems totally enough. Sometimes I accidentally delete an email and then simply hit undo. 
Else I would also say rather desinsitize the action triggered.

tobiasKaminsky:
Undo is already there :+1:

jancborchardt:
Awesome, that's a good conclusion then. :)

korelstar:
Do you have any idea how to desinsitize the swipe feature? I don't know if it's even possible.

However, it seems that the probability to accidentally swipe is very different for different devices.

dan0xii:
> Just let me know some other popular apps which use the same pattern so we can be sure 
> people are reasonably used to it. Otherwise I would say this is pretty strange.

I have seen this in Protonmail, Signal, the reddit app BaconReader and also Telegram to a lesser extent. They all have listviews that vibrate when you long press on an item to indicate that a) you've long pressed and b) there is further action that can be taken.

AndyScherzinger:
We also vibrate on the Nextcloud Android client for long clicks on lists indicating multi select actions. So Nextcloud users are already used to it :)

stefan-niedermann:
I see your point, but long clicking to get more options and swiping which executes one specific option directly are two different things imho.

But since there are more supporters for this topic than the one(s) against it - lets merge it. :)

AndyScherzinger:
> I see your point, but long clicking to get more options and swiping which executes one specific option directly are two different things imho.

That is correct, yes 👍

dan0xii:
I may have mis-understood the original intent.

> In my approach, the vibration is only triggered on movement. 

So if you select an item there is no vibration until you move it up or down in the list?

jancborchardt:
Folks, read the original post. ;) Vibration on long-press is normal, but this is not what this pull request is about.

AndyScherzinger:
> Folks, read the original post. ;) Vibration on long-press is normal, but this is not what this pull request is about.

Sorry, my mistake and you are right: As vibrate isn't used for swipe gestures while a swipe to delete is usually to be combined with a Snackbar offering an undo action. 😃 (one example would be gmail)

stefan-niedermann:
That's correct,amd the undo snackbar is already available. So what about giving a +1 or a -1 to the original post so we can see what the overall opinon is :) that would be great.

AndyScherzinger:
-1 in that case in my personal opinion mimicking the gmail behavior (no vibrate on swipe) seems to be what the users are used to.

stefan-niedermann:
@korelstar i will delete this branch soon. If you want to keep it (e. g. in a private fork), please grab a copy in the next few days.

korelstar:
No problem. I don't need this anymore. Thanks for asking! :-)

stefan-niedermann:
Yep, and while cleaning up, i found this branch: https://github.com/stefan-niedermann/nextcloud-notes/tree/travis-test

What do you wanted to achieve? Do we still need it?

korelstar:
I don't think so. See: https://github.com/stefan-niedermann/nextcloud-notes/pull/188#issuecomment-285469607

PR #211, opened 2017-04-18
Fix #206 Implement splash screen

### LINKED ISSUE #206: Implement splash screen
See https://www.bignerdranch.com/blog/splash-screens-the-right-way/
Material Design Spec: https://material.io/guidelines/patterns/launch-screens.html

stefan-niedermann:
See #206

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -14,14 +14,19 @@
         android:networkSecurityConfig="@xml/network_security_config"
         android:supportsRtl="true"
         android:theme="@style/OwnCloud">
-
+        <activity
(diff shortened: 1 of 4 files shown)


korelstar:
I like this!

What about adding the app name (like a "tagline" in the material design specification) below of the icon (resp. above of the lower border) in a large font but not bold (and IMHO not uppercase like in the documentation example); maybe in blue?

stefan-niedermann:
I thought more about adding an empty material action bar at the top, so it looks like the layout already has been loaded.
Adding a tagline could be nice as well, but i would prefer a grey color.

stefan-niedermann:
It's not as easy as i thought to add some text... i will merge this change as it is already an noteable improvement compared to master and if anyone has time (s)he can make it more beautiful.

Issue #73, opened 2016-02-01
Accept custom certificates


theScrabi:
Hello,

can you please make, that the app is accepting custom certificates I can't install custom certificates since my cert installer on my phone is somehow broken.

stefan-niedermann:
Hmm, it would be better if you could fight the cause, not the symptoms. Have you tried to follow this Guide: https://coderwall.com/p/wv6fpq/add-self-signed-ssl-certificate-to-android-for-browsing

theScrabi:
No I know how to import android certificates xD, its just that the installer is broken due to removal of unnecessary bloatware that afterwards turned out important.

theScrabi:
Its not like you need to fix my errors, I want to switch to lets encrypt eniway, witch will fix that problem. It was just meant as an enhancement, because other owncloud apps also do it.

sedrubal:
I also think it's better to allow several certificates only for several apps and not for the entire system.

So :+1:

theScrabi:
Cool

stefan-niedermann:
Could you please explain me the benefit of allowing a certificate only for an app instead of the whole system? I don't get it. When *you* (as a person) trust a self-signed cert - why should not every of apps (which are all installed by *you*) trust this cert?

sedrubal:
Ok... maybe it's only a very small benefit... :D 

E.g. CACert: I don't want to genreally trust CACert but my owncloud runs with CACert and therefore I want only my owncloud apps to trust CACert.

But then I could only trust the server certificate of my owncloud system wide insteat of CACert root CA... ok...

So I don't know. But the App should at least give a feedback if the certificate is not trusted, but that is #72 I think...

fxedel:
I've just wanted to create this issue, then I saw that it's already there :smile:

Many owncloud installations are on local servers (like mine), so certificates always cause problems. The official owncloud app solves the problem by showing a dialog that the certificate isn't verified and asks if the user wants to continue. That would be my expected behavior: Simple, but the user still has the choice to disagree.

Installing certificates globally is much more complicated (e. g. I have to install Firefox if I only use stock android browser). IMO, there should at least be an error dialog with the message that the certificate isn't verified and a link to an instruction how to solve this.

stefan-niedermann:
> IMO, there should at least be an error dialog with the message that the certificate isn't verified and a link to an instruction how to solve this.

[...]

PR #213, opened 2017-04-30
integrate cert4android

schaarsc:
Proof of concept to solve #73.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "cert4android"]
+	path = cert4android
+	url = https://gitlab.com/bitfireAT/cert4android.git
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -24,6 +24,8 @@ android {
 }
 
 dependencies {
+    compile project(':cert4android')
+
     compile 'com.yydcdut:rxmarkdown:0.1.0'
     compile 'com.android.support:support-v4:25.3.1'
     compile 'com.android.support:appcompat-v7:25.3.1'
--- a/app/src/androidTest/java/it/niedermann/owncloud/notes/util/NotesClientUtilTest.java
+++ b/app/src/androidTest/java/it/niedermann/owncloud/notes/util/NotesClientUtilTest.java
@@ -15,9 +15,9 @@ public class NotesClientUtilTest extends TestCase {
     }
 
     public void testIsValidURLTest() {
-        assertTrue(NotesClientUtil.isValidURL("https://demo.owncloud.org/"));
-        assertFalse(NotesClientUtil.isValidURL("https://www.example.com/"));
-        assertFalse(NotesClientUtil.isValidURL("htp://www.example.com/"));
-        assertFalse(NotesClientUtil.isValidURL(null));
+        assertTrue(NotesClientUtil.isValidURL(null, "https://demo.owncloud.org/"));
+        assertFalse(NotesClientUtil.isValidURL(null, "https://www.example.com/"));
+        assertFalse(NotesClientUtil.isValidURL(null, "htp://www.example.com/"));
+        assertFalse(NotesClientUtil.isValidURL(null, null));
     }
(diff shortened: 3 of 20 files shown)


korelstar:
Wow, that's very cool!

What do you mean exactly with "proof of concept"? What is needed in order to bring this into production state?

stefan-niedermann:
Hey, wow, this would be so great! I just had a look at the code changes, but i am unfortunately unable to test this because i do not have cli-access to my nextcloud instance and i am therefore not able to install a self signed certificate. Maybe anyone (@theScrabi @sedrubal @fxedel @anonimou0) can pull this PR and test it and take some screenshots of the dialog?

Maybe i can find a hour of free time to set up an instance on a local machine...

fxedel:
Does the app still work with ownCloud? If yes, I could test this PR.

stefan-niedermann:
@fxedel as far as i know is the serverside app still completely compatible. the android client does only contact the api of the serverside note app, so it **should** work (though not tested). thanks :)

fxedel:
I get an SSL error:
```
05-02 23:12:19.066 17110-17128/it.niedermann.owncloud.notes E/SyncTask: Exception
javax.net.ssl.SSLHandshakeException: Custom certificate service not available
```
And this should definitely be scrollable (it's even worse in landscape mode):

screenshot_20170502-223154

schaarsc:
@korelstar I called it proof of concept because
 - I did not do any research, alternatives to cert4android maybe available.
 - I changed activity_settings.xml to add some options, just to see if the integration works. In a second step this should be changed to use Android Preference API
 - only tested with server using https

schaarsc:
issues reported by @fxedel should now be resolved.

korelstar:
I finally had some time to have a look at this in more detail. First of all: great work, @schaarsc! Thank you very much for your contribution! Here is some feedback:

- I like the new preference screen. It is much more structured :smiley: 
- The message "Custom certificate service not available" happens quite often (although I used your latest commit). Maybe it's not necessary to create a new instance of `CustomCertManager` on every request? Then you could reuse that instance (maybe as member variable in `NoteServerSyncHelper`?)
- Binding the cert-service should not be done in `NoteSQLiteOpenHelper` since that class is only for database connection. Can you move this to `NoteServerSyncHelper`, please?
- Strings: I suggest to use `Account settings` instead of `server settings`, because the header of this preference section is already `Server`.

schaarsc:
@korelstar I agree code should go into NoteServerSyncHelper and since we bind to the service in this class, we can also hold a reference to CustomCertManager.
I'm also holding a CCM in the SettingsActivity now.
I cannot reproduce the issue with my device. If your are still getting errors cloud you please provide some details on how to reproduce and if possible log-files.

korelstar:
I will test your latest changes and give feedback (with a detailed log) in the next days ...

korelstar:
I just noticed, that the new preferences activity doesn't have a title (ActionBar). It would be nice, if it will get the same title layout like the old settings activity.

stefan-niedermann:
will try to have a look this week

MetroMarv:
I just tried the updated version on my Nexus 5 with latest LineageOS. Everything worked fine until I fetched the updates from my server, by pulling down the notes list view. When I tried first I got an SSL (I don't know the exact exception name) exception and nothing was updated. On my second try I got an IO exception ("unexpected end of stream on Connection(<connection to my server>)"), but the local notes were updated. On my third try everything worked fine.

I didn't had the time to inspect the problem in detail, but I will do this next week.

korelstar:
any news on this, @stefan-niedermann? Did you found some time to test this? Since this affects many files, it would be nice to merge this soon.

stefan-niedermann:
not yet, sorry. i aggree this should get merged soon, i will try.

stefan-niedermann:
merged. thanks for all the hard work and patience.

Issue #217, opened 2017-05-18
Make workflow configurable NotesList -> EditNote -> PreviewNote


ThingsGetWorse:
Hi,
in version 0.11.0 (or possibly earlier, I only got the new feature today via Google Play) PR #141 was merged. I would like to vote to make that flow configurable.

**Options**
    NotesList -> EditNote -> PreviewNote
    NotesList -> PreviewNote -> EditNote

I use the Notes app as an Evernote Replacement so I add notes, sure, but also scroll through and read them. So please make the behavior of the NotesList screen configurable.

stefan-niedermann:
Hi, it is planned to merge EditNote and PreviewNote into one activity with a toggle-button to switch the views. the view will be stored, so of you reopen a note the state will be recovered. would this planned behavior match your request?

ThingsGetWorse:
Hi, that feature sounds good but still requieres me to go through all notes switching state to 'PreviewNote', right?
So request would stay the same, let me chose which  view should be default. :-)

stefan-niedermann:
It is not decided yet if the state will be per note or global - but we can include this to the decision finding :)

ThingsGetWorse:
👍

schaarsc:
I'd like to suggest a preference with three options:
- Always open notes in edit mode (default)
- Always open notes in preview mode
- remember my last decision (for each note???)

minimeee:
I agree with @schaarsc. Just think about the following workflow:

I have some notes that act as "read-only lists" 90% of the time, but sometimes I add or modify something. After doing that, I close the note because I'm done. Next time I'd like to read it, as I'm used to. However, if the app stores the last state, I would immediately enter "edit mode" (which I don't want). In order to get into "view mode" next time I open a note, I'd have to remember to switch to it each time I've edited the note. That's inconvenient imho and not a replacement for a permanent option...

korelstar:
@schaarsc has implemented this in #242 👍

Issue #155, opened 2016-12-12
Merge NoteActivity and EditNoteActivity


stefan-niedermann:
These two Activities should be merged with the aim to use only one activity that is able to toggle between the EditView and the TextView.

Wanted behavior:

**Given** the user is in the EditNoteActivity
**When** the user clicks on the eye-icon
**Then** the EditView should be switched to the TextView
**And** the eye-icon should switch to the [eye-off-icon](https://materialdesignicons.com/icon/eye-off)

**When** the user clicks on the eye-off-icon
**Then** the TextView should be switched to the EditView
**And** the eye-off-icon should switch to the eye-icon

korelstar:
I thought about this, too. An additional way to switch between the two views could be swiping the fragment to the left/right. I also thought about using tabs, but this would consume much space.

stefan-niedermann:
Hm, no Tabs (needs too much space as you recognized yourself). Swiping? I don't know, gestures are very hard to discover and could be triggered accidentely. So *maybe* additional, but the primary way should be the eye/eye-off-icon.

korelstar:
No problem. These were just some thoughts of mine.

korelstar:
Another thought: It is saved whether *edit mode* or *preview mode* was active the last time, and when a (another or the same) note is opened the next time, the Activity switches automatically to the respective state. I envision to save the last state globally, not per note. This would make it more comfortable, when I want to read notes the most time and edit them seldom.

What do you think about this?

stefan-niedermann:
@jancborchardt said
> [...] the edit mode should always be the default.
Though i think saving the state would make sense. *If* we save the state, we should do it globally indeed. Either way we will do it the exact same way like the server-side notes app will do to decrease confusion.
So Saving the toggle state depends on https://github.com/nextcloud/notes/issues/23 and could even be handeled in a seperate issue.

korelstar:
It would be still the default, since for new users, the *edit mode* would be active. I don't know if @jancborchardt really meant that this applies to every *note opening*.

stefan-niedermann:
Some christmas thoughts about how this could be implemented without a performance killer.
I thought we could add one `RxMDTextView` next to the `RxMDEditText` into the `activity_edit.xml`. Then we would have to implement a show/hide toggle mechanism. Problem: When user opens a note and everything runs in the main thread (and TextView *and* EditText are initialized) this could cause a weak performance.
Here is some pseudo-code:

```xml
[...]

PR #221, opened 2017-05-28
Merge activities

schaarsc:
implements #155 #217 

this PR is based on #213 because I wanted to use the new Preferences.

I'd suggest to keep CreateNoteActitvity as separate Activity, because it does not fit well into the new EditNoteActitvity (no auto-save, different ActionBar, different return Intent)

The workflow is implemented as global preference. It should be no problem to add per note option, should  this be required at a later point of time.


stefan-niedermann:
I'll check this in two weeks, i'm on vacation. Thanks for the work.
One hint: It should not be implemented as a preference but simply stored whenever the user toggles the state.

schaarsc:
how do you want to proceed with this PR? This is a big change it already has conflicts and it will only get worse...

korelstar:
@schaarsc Please rebase your branch in order to get rid of the several commits which already has been merged to `master`.

schaarsc:
see #242

PR #230, opened 2017-06-29
Small changes

korelstar:
- [x] remove permission `USE_CREDENTIALS`: I don't think that it is used. But please co-check!
- [x] provide possibility to downgrade the database, this can be useful for development/testing purposes.
- [x] upgrade gradle


stefan-niedermann:
Agree on all changes. USE_CREDENTIALS was used in a early version, i hoped to can reuse the owmCloud/Nextcloud account. Maybe we can introduce this feature (like in the android news app) in the future again.

PR #234, opened 2017-07-31
Auto-save without sync if sync is not possible

korelstar:
I think, saving the note locally from time to time even if sync is not possible is a good idea. This can prevent data loss in some device/os/app failures.


jancborchardt:
Yup, absolutely! :) Can you review @stefan-niedermann?

stefan-niedermann:
Okay, makes definitely sense :) Any idea why this breaks the travia build? I can test tomorrow and will merge if everything is fine :)

korelstar:
Please have a look at the Travis log: there was just a problem while fetching the dependencies. The other test running was fine...

Issue #246, opened 2017-08-31
Edit/Preview Activity enhancements


korelstar:
Thanks a lot to @schaarsc for #242 !

I have some further ideas for enhancement:

- [x] The ActionBar buttons and menu entries should be consistent for edit mode and for preview mode. Currently, there is a "Category" button in edit mode, but none in preview mode. **(high priority)**
- [ ] I suggest a new view mode "source code", which contains a native `EditText` without Markdown functionality. This could be useful, if there are problems with `RxMDEditText` or if someone wants to edit text only. **(low priority)**

Edit: Here are some more issues:
- [x] `NoteEditFragment.getNote()` saves implicitely the notes and triggers synchronization. This is kind of bad hack: E.g., when opening a note in edit mode, `saveData()` is called twice (without a need at all), when triggering the favorite action in edit mode, `saveData()` is also called twice (triggering once at the end is sufficient). This should be optimized. **(high priority)**
- [x] The label of the *edit* button is wrong. A good solution would be to add both (edit and preview) buttons to the XML file and then toggle visibility instead of modifying them. **(high priority)**
- [x] I suggest to change the order of the buttons/menu and put the preview/edit button at first position. In my point of view, this is the most important button which should be shown even on devices which are so small that only one button can be shown. **(medium priority)**
- [x] It's good that the mode strings in `EditNoteActivity` (e.g. `edit`, `preview`, ...) were moved to constants. However, there are still two sources of these strings: `arrays.xml` and `EditNoteActivity`. This should be consolidated into one source. E.g., add constant strings to `arrays.xml` (e.g. `<string name="note_mode_preview">preview</string>`) and reference those from `string-array` (e.g. `<item>@string/note_mode_preview</item>`) and from `EditNoteActivity` (`getString(R.string.note_mode_preview)`). Another option would be to use class constants and then fill the preference activity manually from these constants. **(low priority)**
- [x] We should do this also for all preference keys, such as `PREF_NOTE_MODE` (but also for the other preferences). But that can be done later. **(low priority)**

stefan-niedermann:
I agree with the first, but how would the source code help to fix bugs in Rx?

korelstar:
This is not the purpose. But "live markdown highlighting" will always have a special look&feel that can make things more complicated than editing plain text. It's just my experience from using Notes for a longer time.

But it would be interesting, to get other people's feedback on this.

stefan-niedermann:
cc @jancborchardt

[...]

PR #260, opened 2017-09-08
use R.string for preference key/value

schaarsc:
#246


stefan-niedermann:
May i ask what the benefit of this PR is? Is there any advantage over using java constants?

In my understanding the string ressources were for graphically displayable strings (texts in the gui) rather than for keys?

@korelstar do you know this pattern?

korelstar:
@stefan-niedermann
Still any doubts vor can I merge this?

stefan-niedermann:
Your explaination makes sense, i did not know this pattern before. Merge :)

PR #266, opened 2017-09-18
Update portuguese translation

YSmhXQDd6Z:


korelstar:
Thanks for these translations!

However, the translations are imported from Transifex to this repository. Therefore, please add your translations directly there: https://www.transifex.com/nextcloud/nextcloud/android-notes/

YSmhXQDd6Z:
I know but I always translated this way from the beginning. Can this one be merged as there are no conflicts?

korelstar:
I understand that. But if we would merge this, the changes will we overwritten by the next export from Transifex (this process is automated). Therefore, your work would get lost anyway.

Currently, Transifex is the only way to add new translations. It would be very nice, if you could create a free Transifex account an join the Portuguese language group in order to help translating this app: https://www.transifex.com/nextcloud/nextcloud/

PR #263, opened 2017-09-12
Single note widget scroll

dan0xii:
Allows the single note widget to scroll. A slight drawback is that if the height of the widget is greater than the listview containing the note, there will be a dead space below the note that can't be tapped to open the note -- the user must hit the note text.

PR #268, opened 2017-09-19
Fix to single note widget

dan0xii:
Opening a note from the widget would have no effect if the user already had an editor
open for a different note (mentioned in #263). This change should check to see if a note is already being
edited, save and synchronise it and then present the new note for editing.

Unsure as to whether db.updateNoteAndSync is the best way to immediately save the note before overwriting it with the new one.

### DIFF
--- a/app/src/main/java/it/niedermann/owncloud/notes/android/activity/EditNoteActivity.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/android/activity/EditNoteActivity.java
@@ -75,6 +75,26 @@ public class EditNoteActivity extends AppCompatActivity implements CategoryDialo
         }
     }
 
+    @Override
+    protected void onNewIntent(Intent intent) {
+        super.onNewIntent(intent);
+        setIntent(intent);
+    }
+
+    @Override
+    protected void onResume() {
+        super.onResume();
+        Intent intent = getIntent();
+        DBNote note = (DBNote) intent.getSerializableExtra(PARAM_NOTE);
+
+        // Does the note retrieved from the intent match the note currently
+        // being displayed by the app
+        if ((note.getId()) != fragment.getNote().getId()) {
+            originalNote = note;
+            createEditFragment(note);
+        }
+    }
+
     private void createEditFragment(DBNote note) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/it/niedermann/owncloud/notes/android/activity/EditNoteActivity.java
```diff
@@ -75,6 +75,27 @@ protected void onCreate(final Bundle savedInstanceState) {
...
+
+    @Override
+    protected void onResume() {
+        Intent intent = getIntent();
+        DBNote note = (DBNote) intent.getSerializableExtra(PARAM_NOTE);
+
+        // Does the note retrieved from the intent match the note currently
+        // being displayed by the app
+        if ((note.getId()) != fragment.getNote().getId()) {
+            db.updateNoteAndSync(fragment.getNote(), fragment.getNote().getContent(), null);
```
korelstar: I don't think that you need this, since your `NoteEditFragment.getNote()` saves the note automatically, when called (I don't like this, but that's how it's currently works).

app/src/main/java/it/niedermann/owncloud/notes/android/activity/EditNoteActivity.java
```diff
@@ -75,6 +75,27 @@ protected void onCreate(final Bundle savedInstanceState) {
...
+            originalNote = note;
+            createEditFragment(note);
+        }
+        super.onResume();
```
korelstar: Is there a need, that the superclass method is called at the end? The [documentation](https://developer.android.com/guide/components/activities/activity-lifecycle.html) says: "Always call the superclass first"

dan0xii:
> I don't think that you need this, since your NoteEditFragment.getNote() saves the note automatically, when called (I don't like this, but that's how it's currently works).

I didn't spot that, thanks. 

> Is there a need, that the superclass method is called at the end? The documentation says: "Always call the superclass first"

That's a mistake, I wasn't aware of that. I've fixed both now and updated the branch.

korelstar:
Thanks!

PR #270, opened 2017-09-26
Design of note list item

korelstar:
improve design (remove 1dp from the padding of the new category box, because that took too much space) and code (fixed some lint warnings).


stefan-niedermann:
I wasn't sure, i didn't want to make the box looking too slim. But i am okay with decreasing padding for now :)

PR #273, opened 2017-10-08
Note list widget

### LINKED ISSUE #15: Make a widget for all Notes
Hey @stefan-niedermann, cool app.

Would be awesome if you could enhance this app with a widget. Then I would start using it as my new standard app for Notes. :)

dan0xii:
Widget to display all notes in the database, sorted by starred,
modified. updateNoteListWidgets() used temporarily for widget updates until
content observer is implemented. Closes #15.

Removed PARAM_WIDGET_SRC from EditNoteActivity because it was no longer used.

Small clean up of SingleNoteWidget and CreateNoteWidget.


dan0xii:
Sorted. Yes, I have tested it but could do with someone else's feedback. I'm going to try and implement a content provider/observer next.

stefan-niedermann:
I agree that the star should be right. It simply wastes too much space and if nothing is starred it looks odd. *And* i have seen recent issues at the nextcloud server repo: they are removing the star on the left side (at least on mobile), and use the star only as a badge on the file icon in the future. this gives us freedom where to put it since we do not have icons.

tl;dr put the ★ on the right side

dan0xii:
I've moved the star icon so you can at least see what that looks like. Much better imo.

I'll look at point 1) raised by @korelstar above.

> Sometimes, an empty item is shown in the list. This happens only, if a note's title is longer than the widget width. You should add android:lines="1" and android:ellipsize="end" to the layout file (for widget_entry_content_tv). However, this doesn't fix the problem completely. Maybe, you can find the problem (if you can reproduce this)?

I haven't seen this in the time I've been using it and I have note titles that are either a long collection of words (i.e. separated by whitespace) or URLs and in both cases they are just cut off at the end of the widget entry. What happens if you tap the empty entry, does it open the note as expected?

Thanks.

korelstar:
- The star still consumes space, even if the note is not starred. You should set the visibility of the `ImageView` to `GONE`, if the note is not starred. Additionally, you should set the image source in the XML and not in Java code (the source is not suspect of any change).

- A title with long words like "Die Donaudampfschifffartsgesellschaft" (German has many long words) is currently shown as "Die", since the whole second word is hidden. If you use `ellipsize` and `lines` as proposed above, the title will be shown as "Die Donaudampfs.." which is much more usable.

- I can reproduce the error (not displaying the title) with the following steps (this works on my phone only; and not in the emulator):
  1. Scroll the list down, so that an arbitrary note is at the top margin and already half hidden. You can see only the bottom half of the note's title.
  2. Tap on the `+` to start the new note activity.
  3. Close the activity using `X`
  4. The title of the half shown note is not visible anymore, although the list hasn't moved.
  5. You can scroll up, as you would do to see the whole title of the respective note; but the title isn't shown anymore.
  6. additionally, if I scroll down or up, the title of some other notes are not shown anymore.
  7. If I click on an empty title, the correct note is opened.

dan0xii:
Hold off on reviewing this widget until I finish testing. I still have a couple of things to do.

dan0xii:
1) I've made the requested change to widget_entry and so long words are abbreviated properly but I can't reproduce an emtpy item being shown either on my phone or in the emulator.

2) I can't reproduce this either. How many notes do you have? I've got around 50 after generating some random ones.

3) Done.

korelstar:
1. Thanks, but the bug is still there (see below).
2. I have more than 70 notes, in summary about 90 KiB. However, the loading view is visible much more seldom than before. I think, this is sufficient for now.
3. Thanks. :-)

ad 1: I made a video of the problem: http://daten.korelstar.de/unsortiert/notes-widget.mp4
I added 9 test notes before. In the video, I scroll to show these notes. Then I press `+`, the create note activity starts, I click on `X` to cancel. Now, note "Test 4" is not shown anymore. If I would click on the space between "Test 3" and "Test 5", the note "Test 4" is opened correctly (not shown in the video). Then I repeat the whole thing and "Test 2" is not shown anymore. When I scroll down, "Test 9" is also not shown anymore.

@stefan-niedermann How is the behavior on your phone?

dan0xii:
That video is really helpful. I've not seen anything like that but I'll look into it.

I recall that you have a Wileyfox Swift but could you tell me what version of Android you're running?

I'll setup an emulator with the same version and increase the number of notes I have in case that makes a difference.

korelstar:
I'm using CyanogenOS 13.1.5 / Android 6.0.1.

Thanks for investigating this. I tried this, too, but I wasn't successful.

dan0xii:
@korelstar would you re-test for me. I can't reproduce the problem but from reading the docs I think
I wasn't setting the listview's emptyview correctly. I vaguely remember thinking I would come back
and fix it and clearly didn't. I'm not 100% sure this is the cause but I can't see clues anywhere else.

korelstar:
Cool! It works, now! What exactly did you changed?

From my side, this PR can be merged, now. AndroidStudio still gives some warnings/suggestions (e.g. unused import statements, object fields that can be converted to local variable, better support for right-to-left-layouts (add e.g. paddingStart when paddingLeft exists)), but I'm fine with solving those later (I don't want to increase the probability of conflicts to my big open PR 😁).

dan0xii:
Great, good news. The listview allows you to set an "empty" view should there be no items to display in the list and I hadn't implemented it properly. The only scenario where I could see this happening would be if the user logs into their account, places the widget and then deletes all the notes.

I've removed the two un-used imports from NoteListWidgetFactory (which shouldn't cause you any grief :)) and I'll create a separate PR for the other warnings that you mention.

PR #281, opened 2017-10-17
[tx-robot] updated from transifex

korelstar:
(can be left open until the next release is arranged)


stefan-niedermann:
Oops, forgot this for `0.13.0`.

korelstar:
You also forgot to set the new version in `app/build.gradle`! Now, we have a tag `v0.13.0` with version name `0.12.1`. Can this be a problem with the automatic releases on F-Droid?

I suggest to go directly to `0.13.1` (tag and `build.gradle`) in order to avoid problems.

PR #297, opened 2017-11-11
Reworked single note widget

dan0xii:
The widget will now check to see if the note exists before trying to display it. Fixes #293.
* Moved opening of db from getViewAt.
* Fixed the widget's empty view.

PR #298, opened 2017-11-13
Widget cleanup

dan0xii:
General improvements to widget code

* Removed unused imports
* Improvements to accessibility and internationalisation
* Removed hard coded dimensions in layouts
* Set the note list widget's empty view to be displayed in the scenario when the user adds the widget to their home screen then proceeds to delete all notes from within the app.

This includes the changes from #297 so it will be worth waiting until that's agreed before reviewing this PR.


app/src/main/java/it/niedermann/owncloud/notes/model/SingleNoteWidgetFactory.java
```diff
@@ -48,57 +54,56 @@ public int getViewTypeCount() {
...
     public void onDataSetChanged() {
+        SharedPreferences sharedprefs = PreferenceManager.getDefaultSharedPreferences(mContext);
+        long noteID = sharedprefs.getLong(SingleNoteWidget.WIDGET_KEY + mAppWidgetId, -1);
 
+        if (noteID >= 0) {
+            if (db.getNote(noteID) == null) {
+                Log.e(TAG, "Error: note not found");
+                note = null;
+            } else {
+                note = db.getNote(noteID);
```
korelstar: If the note exists (which is the default scenario) `db.getNote(noteID)` is executed twice. Since database queries are expensive, this should be optimized.

app/src/main/java/it/niedermann/owncloud/notes/model/SingleNoteWidgetFactory.java
```diff
@@ -48,57 +54,56 @@ public int getViewTypeCount() {
...
+            }
+        }
     }
 
     @Override
     public void onDestroy() {
-
+        db.close();
```
korelstar: Since `NoteSQLiteOpenHelper` is currently a singleton, it's dangerous to close the database from here. Theoretically, it's possible that another objects still wants to use the database. Hence, I suggest to remove this.

app/src/main/res/layout/widget_entry.xml
```diff
@@ -4,29 +4,34 @@
     android:layout_height="wrap_content"
     android:layout_width="match_parent"
     android:orientation="horizontal"
-    android:padding="4dp">
+    android:padding="@dimen/widget_note_list_entry_layout_padding">
```
korelstar: What is the advantage of moving these dimensions to `dimens.xml`? As far as I can see, you are not re-using any of them! From my point of view, this makes everything much more complex ...

app/src/main/res/layout/widget_entry.xml
```diff
@@ -4,29 +4,34 @@
...
 
     <TextView
         android:id="@+id/widget_entry_content_tv"
-        android:layout_width="wrap_content"
+        android:layout_width="@dimen/widget_note_list_entry_content_width"
         android:layout_height="match_parent"
-        android:paddingLeft="6dp"
+        android:paddingStart="@dimen/widget_note_list_entry_padding_start"
+        android:paddingEnd="@dimen/widget_note_list_entry_padding_end"
```
korelstar: Why do use different dimensions for padding left and start (respective right and end)? These values should be the same and therefore the dimension keys, too (if used at all, see other comment). I don't think, that the file `values-v17/dimens.xml` is needed.

dan0xii:
Hello Korelstar,

I was intending on using mAppWidgetId in NotesListWidgetFactory but have removed it for now because it will be a while before I get around to it.

The TODO for the loading view is a reminder but I will remove it if you prefer. I noticed that the "Loading" text isn't formatted great but it only appears on my phone for a split second every so often.

I'll work on the getNote() operation in SingleNoteWidgetFactory. I was trying to avoid using try-catch. I'll remove the call to db.close() also, that hadn't occured to me.

I moved the dimensions to a separate file as I was under the impression it's best practice to allow for different layout sizes on different devices (with varying display sizes).

I agree I could have used the same key name for the left/right and start/end dimensions but I thought it might be confusing to have `paddingStart="@dimen/padding_left"`. The file `values-v17/dimens.xml` is needed to provide proper layout support for right-to-left devices. Once we are only supporting API 17 and upwards paddingLeft and paddingRight will be ignored when paddingStart and paddingEnd exist.

korelstar:
> The TODO for the loading view is a reminder but I will remove it if you prefer.

No, that's fine.

> I'll work on the getNote() operation in SingleNoteWidgetFactory. I was trying to avoid using try-catch.

Just save the result as local variable.


> I moved the dimensions to a separate file as I was under the impression it's best practice to allow for different layout sizes on different devices (with varying display sizes).

Okay, but I still think, that maintainability would be increased if you use one single dimension-key for all usages that belong to each other. E.g., introduce a dimension `widget_note_list_outer_padding` with value `6dp` an use this one instead of `widget_note_list_header_icon_padding_left`, `widget_note_list_header_icon_padding_start`, `widget_note_list_create_icon_padding_right`, `widget_note_list_create_icon_padding_end`, `widget_note_list_title_padding_left`, `widget_note_list_title_padding_start`, `widget_note_list_entry_padding_left`, `widget_note_list_entry_padding_start`, `widget_note_list_fav_icon_padding_right` and `widget_note_list_fav_icon_padding_end`. Motivation: If you change one of them, all of them have to be changed to the same value. Otherwise, the widget will look strange.

> I agree I could have used the same key name for the left/right and start/end dimensions but I thought it might be confusing to have `paddingStart="@dimen/padding_left"`.

If you name it `padding_start` it looks better. Even for `paddingLeft="@dimen/padding_start"`, since `paddingLeft` is deprecated.

> The file `values-v17/dimens.xml` is needed to provide proper layout support for right-to-left devices. 

I don't understand this. If you use `padding_start` (or something like `widget_note_list_outer_padding`, see above) for both `paddingLeft` and `paddingStart`, then the definition must be in `values/dimens.xml` and you don't need `values-v17/dimens.xml` anymore. Nevertheless, right-to-left is fully supported this way.

> Once we are only supporting API 17 and upwards paddingLeft and paddingRight will be ignored when paddingStart and paddingEnd exist.

Isn't this already the case when the app is executed on a device supporting the respective API?

dan0xii:
Take a look at the latest revision. I think I have ironed out all of the issues.

stefan-niedermann:
Nope i'm fine with the changes, merge :)

PR #304, opened 2017-12-05
Dark theme

dan0xii:
Configured night mode to allow for a dark theme within the app.


korelstar:
Thanks for your PR! Finally, I had some time in the train to look at this. Here are my remarks:
- In `PreferencesFragment.onPreferencesClick`, the string `Dark` has to be moved to `strings.xml`.
- The night color for `bg_attention` should be adjusted, e.g. to `#ff3333` (this is better, but not good).
- I think, we should ask @jancborchardt resp. @nextcloud/designers for their opinion about the chosen colors.
- I'm not sure, but I would suggest to remove colors from `night/colors.xml` which are not changed. But other opinions are welcome.
- Another topic to discuss: Android provides four settings for the night mode: `MODE_NIGHT_NO` (always use day theme), `MODE_NIGHT_YES` (always use night theme), `MODE_NIGHT_FOLLOW_SYSTEM` (use system setting for night mode, this should be the default) and `MODE_NIGHT_AUTO` (chose theme in dependency of the current time). I suggest to provide all four values in the settings. Currently, you only provide `YES` and `NO`. Any other opinions?

dan0xii:
> I'm not sure, but I would suggest to remove colors from night/colors.xml which are not changed. But other opinions are welcome.

My memory is not 100% clear on this but I'm pretty sure that I didn't include them at first and I would find elements displaying the wrong colour, i.e. not what you would think based on the "day" colours, and so I added them in to fix it. The single note widget's text colour springs to mind.

> Another topic to discuss: Android provides four settings for the night mode: MODE_NIGHT_NO (always use day theme), MODE_NIGHT_YES (always use night theme), MODE_NIGHT_FOLLOW_SYSTEM (use system setting for night mode, this should be the default) and MODE_NIGHT_AUTO (chose theme in dependency of the current time). I suggest to provide all four values in the settings. Currently, you only provide YES and NO. Any other opinions?

I wouldn't personally use a feature to have the app detect day or not and change the theme appropriately but if there is a need for this I would build it in. 

I'll take a look at the other points you raised.

korelstar:
to doubled colors: okay, then don't touch this.

to night mode settings: what do you think, @stefan-niedermann ?

stefan-niedermann:
I would go for YES, NO and FOLLOW_SYSTEM and consider the forth option only if a user requests it.

dan0xii:
This is ready for review. There are still only "Light" and "Dark" options (YES/NO) because I can't get FOLLOW_SYSTEM to make any difference. "Automatic" works fine.

I've fixed the repeated colours and adjusted the shade of bg_attention. Left to do is agree on the colours to be used.

I'm going to implement a pure black "Dark" theme in a separate PR once this is finalised.

I don't know why the Travis build fails btw--it's fine on my machine and all the SDK licenses are accepted.

stefan-niedermann:
@dan0xii can you please rebase and resolve the conflicts with current master? I'd like to get this PR in :)

korelstar:
It appears that LineageOS provides an OS global switch for the night mode, see:
https://www.lineageos.org/Introducing-the-LineageSDK/
https://wiki.lineageos.org/sdk/api/styles
Can we support this, too? It's compatible to non-LineageOS devices.

jancborchardt:
For a design review, I need to see some screenshots :)

dan0xii:
@stefan-niedermann I've just rebased it.

stefan-niedermann:
Great, i will have a look as soon as possible - but today i don't want to see the app anymore :laugh:

jancborchardt:
@dan0xii can you post some screenshots? :)

jancborchardt:
Oooh that looks very nice! :) Some feedback:

- [ ] We can't use the blue for font color on dark in the sidebar for the active entry as it's not visible (and especially less visible than the inactive entries). Let's use white for the active one, and white with 70% opacity for the other ones.
- [ ] The background is not consistently the same shade of grey. In the notes detail view it seems full black (too harsh), in the notes list it's also black, but on the bottom (below the last note) it's lighter. And the theme chooser modal is even lighter. In general it's best to not use real black as that's very harsh.

Agree with @stefan-niedermann we should simplify it to just a switch saying "Use dark theme".

dan0xii:
Take a look now. I've gone back to using a switch to change between light and dark theme as before. The colours should be more consistent but would appreciate a second pair of eyes to check it over. Here's what it looks like:

screenshot_1523825431
screenshot_1523825436
screenshot_1523824086
screenshot_1523824113
screenshot_1523823881
screenshot_1523824098

stefan-niedermann:
Okay, i merged to master. We can continue polishing it there, but i think we already have a good standing :-)

Thanks @dan0xii !

PR #198, opened 2017-03-27
[Work in Progress] Sidebar

stefan-niedermann:
- [x] Implement sidebar with basic entries `All Notes` and `Favorites`
- [ ] Add hamburger-button to the ActionBar
- [ ] Make entries functional
- [ ] Move `About` to sidebar
- [ ] Move `Settings` to sidebar as a bottom point
- [ ] Add some nice icons to each point

Displaying categories in the sidebar will be a separate step.

device-2017-03-27-231604

PR #275, opened 2017-10-15
Categories in NavigationDrawer

korelstar:
This work in ready for merge, now. Closes #165.

- [x] introduce NavigationDrawer with hamburger icon
- [x] some code improvements/refactoring in `NotesListViewActivity`
- [x] move "settings" and "about" from menu to navigation
- [x] show "current" and "starred" in navigation
- [x] show every existent category in navigation
- [x] app is started with selection "current" (same behavior like before)
- [x] filter notes list in depdendeny of navigation selection
- [x] show notes in alphabetical order if category is selected
- [x] don't show starred notes above of not-starred notes (because there is the new selection "starred")
- [x] highlight current selection in navigation
- [x] set ActionBar title to indicate current category selection (?)
- [x] add icons for each navigation entry
- [x] show notes count for each navigation entry
- [x] load category list in background thread
- [x] refresh category list if notes list is refreshed
- [x] only show first-level categories in navigation / show notes of all sub-categories in notes list
- [x] show sub-category as header in notes list
- [x] set category category of new note to the currently selected category, when clicking on the create note button
- [x] keep current category selection when rotating device
- [x] don't show category in list of notes items, if a category is selected (?)
- [x] show sub-categories (second level) by clicking on the icon in the navigation (?)
- [x] check (and fix?) compatibility to note selection for single note widget
- [x] check (and fix?) indentation of changed files (4 spaces vs. tabs)
- [x] replace "/" by " / " in category titles everywhere in the GUI (this makes reading of sub-categories easier)
- [x] provide a drop-down list of existing categories in the change-category dialog
- [x] don't update changed-time of the note when updating the category (since it's meta-data, same behavior when starring a note)

PR #307, opened 2017-12-16
Drawer menu implementation

AndyScherzinger:
This PR is a new take (compared to #198) on implementing the drawer menu (basically what I implemented on the Nextcloud client and Nextcloud SMS client itself). 
Due to my IDE setup this PR will also update the build tools used, the gradle wrapper and the Android gradle plugin version too. So this PR needs Android Studio 3.0.1 to work. If this is a problem than I can revert back the files that lead to this (their are not necessary for this PR but have been auto-updated by Android Studio 3.0.1).

This PR adds a Drawer Implementation to the main view of the app adding 
* 3 menu items via a menu configuration file
* adding an abstract  DrawerActivity which takes care of the drawer handling (screen rotation state preservation, menu icon tinting, etc.)
* it doesn't add a "favorites" item since I haven't seen this functionality yet but that would be easy to do - once the functions is available one just needs to add another menu items + action.

Here are 2 screenshot of how it looks on Android 8 (and since it is implemented with the standard appCompat classes it also looks like this on older Android versions :)

device-2017-12-16-180827
device-2017-12-16-180838

Please let me know if anything needs to be changed and don't hesitate to ask any questions that might come up.

cc @jancborchardt @korelstar @stefan-niedermann 😀 

Hope you like my work ❤️


AndyScherzinger:
Pinging @korelstar @stefan-niedermann for review / feedback :)

korelstar:
Sorry, this month (and I think also the next one) I have no time for this project. But after that, I'll be back ;-)

Nevertheless, I've noticed your PR and I thank you very much for your contribution. However, I'm very confused about it, since we've introduced the NavigationDrawer already in #275 (it was merged into `master` some time ago (before your PR) but there isn't any new release with it yet). Therefore, #198 is obsolete and should be closed. Hence, I think your work is also already obsolete (I assume that you hadn't seen the process in `master`) -- I'm very sorry to say that. On the other hand, maybe there are some aspects of your implementation that we should integrate into the current solution?

AndyScherzinger:
@korelstar thanks for the feedback. I missed #275 and only saw #198. Looking at #275 it seems to me that one already implements anything which this PR would have added. So it can likely be closed.

PR #324, opened 2018-02-07
Moved widget update functions 

dan0xii:
Moved widget update functions from NoteServerSyncHelper to NoteSQLLiteOpenHelper to allow the widgets to be updated without having to wait for a sync with the server.

Currently there is a bug where you can crash the app after deleting a note via the Note List Widget. Steps to reproduce:

1. Select a note from the Note List Widget so that you're presented with the edit note window.
2. Go to the context menu and delete the note, you'll be dropped back to your home screen.
3. The Note List Widget still displays the note in its list that you've just deleted.
4. Tap it again and you get a crash.

### DIFF
--- a/app/src/main/java/it/niedermann/owncloud/notes/persistence/NoteServerSyncHelper.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/persistence/NoteServerSyncHelper.java
@@ -38,8 +38,6 @@ import it.niedermann.owncloud.notes.util.NotesClientUtil.LoginStatus;
 import it.niedermann.owncloud.notes.util.ServerResponse;
 import it.niedermann.owncloud.notes.util.SupportUtil;
 
-import static android.content.ContentValues.TAG;
-
 /**
  * Helps to synchronize the Database to the Server.
  */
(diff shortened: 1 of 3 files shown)


app/src/main/java/it/niedermann/owncloud/notes/persistence/NoteServerSyncHelper.java
```diff
@@ -395,9 +393,6 @@ protected void onPostExecute(LoginStatus status) {
             for (ICallback callback : callbacks) {
                 callback.onFinish();
             }
-            Log.d(TAG, "onPostExecute: Update widgets");
-            dbHelper.updateSingleNoteWidgets();
-            dbHelper.updateNoteListWidgets();
```
korelstar: I think this shouldn't be removed. When you receive changes from the server (new note, deleted notes or updated notes), the widgets should still be updated.

dan0xii:
I've reverted the change to NoteServerSyncHelper but I've had to add the widget update functions to the addNoteAndSync and updateNoteAndSync functions. There is still a bug on really slow devices where you delete a note and it is still displayed in the note list widget but I can very rarely replicate it.

korelstar:
Did the changed by myself, since I wanted to get this into #327.

dan0xii:
Great. I didn't spot the duplication.

PR #326, opened 2018-02-11
update dependencies (android, support libraries, cert4android)

korelstar:
Some dependency updates for the next major release.

The "Create note" notification (see #142) uses `NotificationCompat.Builder` in a way which is deprecated now. This should be fixed. Otherwise, we could remove the notification thing, since it's not a widespread UI pattern and, in the meanwhile, a shortcut for the home screen was added (see #197).

Updating to Android 27 should also fix #326.


stefan-niedermann:
@kas70 are the home screen short cuts (#197) an appropriate option for you? I honestly would like to get rid of the notification completely as i already mentioned in #142 that i do not feel comfortable with this since i have doubt it is wildely used.

stefan-niedermann:
#348 will allow us to remove the permanent notification in favor of a new quick tile.

kas70:
@stefan-niedermann sorry for not answering your comment. I'm perfectly fine with this :)

PR #327, opened 2018-02-11
release v0.13.1

korelstar:
`app/build.gradle` wasn't updated before tagging version `v0.13.0`! Now, we have a tag `v0.13.0` with version name `0.12.1`. Can this be a problem with the automatic releases on F-Droid?

I suggest to go directly to `0.13.1` (tag and `build.gradle`) in order to avoid problems.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -7,8 +7,8 @@ android {
         applicationId "it.niedermann.owncloud.notes"
         minSdkVersion 14
         targetSdkVersion 26
-        versionCode 18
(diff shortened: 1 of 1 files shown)


korelstar:
@stefan-niedermann Are you fine with this procedure (merge this and tag `v0.13.1`)?

stefan-niedermann:
or i remove the v0.13.0 tag since the latest build was not available on f-droid

stefan-niedermann:
meh, probably better to jump to v0.13.1.

Issue #344, opened 2018-03-09
»Syncing« icon looks like a warning


jancborchardt:
The current syncing icon looks a bit like there’s an error, because it uses the icon for »sync problem«:
![screenshot from 2018-03-09 16-24-22](https://user-images.githubusercontent.com/925062/37214828-8336b55e-23b6-11e8-8f29-4f8626018963.png)
There is no problem though :) so we should simply use the icon for »sync«: https://material.io/icons/#ic_sync

cc @stefan-niedermann @korelstar

korelstar:
Currently, the icon doesn't indicate that the note is syncing, but it indicates that the note wasn't synced yet (please try offline mode to check). Therefore, I think the icon is appropriate.

jancborchardt:
Something not being synced yet is _not_ an error. :)

In the desktop client for example, a changed file is shown with the blue sync icon while it is not synced. Not an error icon.

korelstar:
Okay, we could try it ...

stefan-niedermann:
Okay, we are using the icon in blue without the exclamation mark for now.

I tested the  icon from the nextcloud android app, too, but it was too attention begging and collapsed with the category. (https://raw.githubusercontent.com/nextcloud/android/master/src/main/res/drawable-xxhdpi/ic_synchronizing.png)

![grafik](https://user-images.githubusercontent.com/4741199/38558078-5d3703c2-3ccf-11e8-83e1-bd36b039a309.png)

PR #345, opened 2018-03-11
update App Bar labels

### LINKED ISSUE #334: Simplify header text
The header text only needs to be one line, the two lines are not really necessary:

- "Notes" (in sidebar "All notes" instead of "Recent")
- "Favorites" (keep consistent with Files/server wording, same for sidebar)
- "Uncategorized"
- "Settings"

korelstar:
Fixes  #334.

- [x] use titles instead of subtitles in `NotesListViewActivity`
- [x] "Notes" (in sidebar "All notes" instead of "Recent")
- [x] "Favorites" (keep consistent with Files/server wording, same for sidebar)
- [x] "Settings" and "Account settings"

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -45,13 +45,14 @@
 
         <activity
             android:name="it.niedermann.owncloud.notes.android.activity.SettingsActivity"
-            android:label="@string/app_name"
+            android:label="@string/settings_server_settings"
+            android:parentActivityName="it.niedermann.owncloud.notes.android.activity.PreferencesActivity"
             android:windowSoftInputMode="stateVisible"
             />
 
         <activity
             android:name="it.niedermann.owncloud.notes.android.activity.PreferencesActivity"
-            android:label="@string/app_name"
+            android:label="@string/action_settings"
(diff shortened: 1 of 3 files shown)


stefan-niedermann:
Nice :) i wanted to change the settings icon so it matches the one from the nextcloud app, gmail, etc (the single gear-wheel). I think we could do this within this PR-Request.

This might be a bit off-topic, but do you know since which API-Version the Vector Drawables were available completely? I would like to drop the  raster graphics completely.

korelstar:
It's Android 5.0, so we should keep the bitmap drawables.

If you are working on icons, you could also replace the sync icon, see #344. But please note, that it is smaller!

stefan-niedermann:
I'll do the icons on a separate branch.

Side note: 3,6% of Play Store customers are using Android < 5.0. Of course we do not have statistics for F-Droid, but i think when Play Store lowers at 0,1%, we can rethink about increasing the API level.

I've also heard that Android Studio is capable of generating the raster graphics from the vector drawables at built time, which could be another option to get rid of the raster graphics but keep support for Android < 5.0

jancborchardt:
Thanks a lot @korelstar! :)

PR #347, opened 2018-03-12
Single note widget minimum height changed from two rows to one.

### LINKED ISSUE #343: Single note widget: possibility to reduce height to 1 row
Currently the minimum height of the single note widget is 2 rows. It would be great to be able to reduce it to 1 row to have a 1 height full width note display.

dan0xii:
As requested in #343.

Closes #343


jancborchardt:
Thank you so much @dan0xii! :) Sorry for the late thanks :tada:

PR #348, opened 2018-03-14
Added support for a quick settings tile for creating a new note

Dobli:
This adds support for adding a quick settings tile on devices running Android 7.0 (API24) and up.

The tile opens the new note activity when clicked (behaving like the [persistent notification](https://github.com/stefan-niedermann/nextcloud-notes/pull/142)) for creating quick notes. 
I works with the screen unlocked as well as from the lockscreen (it asks for pattern / pin / fingerprint when locked). When no account is setup yet it prompts to add one first.

Here are screenshots having it in the quick settings bar and page:

qs_new_note
qs_new_note_open

I tested it with a LG V30 (Android 7.1), so tests with other API version may be necessary (although it shouldn't be visible on devices below API24).

The displayed label reuses the new note string so additional translations are not necessary for this. The icon used is a white version of the launcher icon, but it could be replaced by another one (or by a svg version).


Dobli:
I just noticed another open PR to [merge create and edit activities](https://github.com/stefan-niedermann/nextcloud-notes/pull/342), my solution would need a little change to be working with that PR (essentially only changing the intent to call `EditNoteActivity` instead of the removed `CreateNoteActivity`), so that probably should be considered first.

stefan-niedermann:
Great idea! This looks like an appropreate replacement for the permanent notification :)

I will merge #342 first - it would be cool if you could adapt this pull request then :)

Dobli:
Sorry for the delay.
I now have adjusted the tile to use the merged EditNoteActivity.

PR #386, opened 2018-04-17
Configuration screen for Note List Widget

dan0xii:
When the user places the Note List Widget on their home screen this change allows them to choose whether the widget will show all notes, starred notes or only notes in a specific category.

Screenshot:
screenshot_1523992168


app/src/main/res/values/strings.xml
```diff
@@ -126,6 +126,10 @@
     <string name="widget_entry_fav_contentDescription">Star icon is used to denote an item as a favourite</string>
     <string name="widget_app_launcher_contentDescription">Launches app</string>
 
+    <string name="nlw_display_all">All</string>
```
stefan-niedermann: Use `label_all_notes`

app/src/main/res/values/strings.xml
```diff
@@ -126,6 +126,10 @@
...
+    <string name="nlw_display_starred">Starred</string>
```
stefan-niedermann: use `label_favorites`

app/src/main/res/layout/activity_note_list_configuration.xml
```diff
@@ -0,0 +1,85 @@
...
+            android:gravity="center|bottom">
+
+            <Button
+                android:id="@+id/nlw_done"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:layout_marginBottom="10pt"
+                android:shadowColor="@color/fg_default_low"
+                android:text="Done" />
+        </LinearLayout>
```
korelstar: The four labels in this layout should use (existing) localized strings, too.

dan0xii:
I'll amend those strings, frustrating that I missed them. Here are some screenshots:
screenshot_20180418-151012
screenshot_20180418-151016
screenshot_20180418-151023

stefan-niedermann:
Thanks :)

Can we use radio buttons for the "Display" select? Because we have currently lots of space and we could use it to display the three options directly (ideally with star, folder and clock icons on the left side).

To be honest i am not happy with the buttons at the bottom, too. Maybe each button should be 50% width, the cancel button should not be primary blur but secondary grey, because it is not as important as the submit button.

korelstar:
I suggest to remove all buttons (also no floating button) and provide a list like in the NavigationDrawer:

- All notes
- favorites
- category 1
- category 2
- ...

If you tap on an entry, you chose it. No need for an OK buton, no need for cancel (since it's done using "back").

korelstar:
I suggest to remove all buttons (also no floating button) and provide a list like in the NavigationDrawer:

- All notes
- favorites
- category 1
- category 2
- ...

If you tap on an entry, you chose it. No need for an OK buton, no need for cancel (since it's done using "back").

stefan-niedermann:
That's it, like @korelstar said :) Sounds good. And like in the navigation drawer, we should display the same icon at the left side.

dan0xii:
My intention is to add an option for a dark theme but I held off until it was finalised in the main app.

If we put all the options available on display at once with a single choice I won't be able to add that second option for theme.

I'll mock up a couple of versions so we can see better how it would look.

korelstar:
You could place the dark theme switch above or below of the proposed list:

LinearLayout
- ListView (if required in a ScrollView, but I don't think that this is needed)
- Switch

The default for the switch should be the app setting for dark mode.

stefan-niedermann:
@korelstar you are suggesting the last "entry" (the theme switch) should be sricky at the bottom?

img_20180419_070334

dan0xii:
@stefan-niedermann, would there then be an OK button at the button right hand corner (possibly on a new row) in your mock-up? The design that korelstar was suggesting would mean the user would select an item from the list (all, favourites, category) and that would close the activity.

stefan-niedermann:
No, in my mock up is no submit button, like @korelstar suggested: slecting an entry would immediately close the activity.

The only "problem" might be, that a user selects something, and wanted the other theme, but 

1. most people will use the theme they are already using for the app
2. in case someone did accidently chose the wrong, he can easily remove an readd thw widget
3. this type of widget is not often used i guess - it is (in oppossite to the single note widget) only once or only once per category used, so 2. would not happen very often

dan0xii:
> The only "problem" might be, that a user selects something, and wanted the other theme,
[...]

I see. I think you're right it's a pretty uncommon case. I'll work on the new design and we can see what it's like in practice.

korelstar:
If the default widget theme is the app theme, it's even possible to put the option in a three-dot-menu as a checkbox.

dan0xii:
This is what I have now:
screenshot_1524431289

You make a selection and the appwidget is created. I've taken out the functionality for the dark theme for now because it was complicating matters quite a bit. 

The configuration activity will be shown in the dark theme if the user has selected that in the main app but the widget will only ever be "light" which may be confusing to some users. The dark theme for this activity needs the font colour correcting but looks like this so far:
screenshot_1524432689

stefan-niedermann:
So the widget itself will not be able to be displayed in dark theme (yet)?

The config screen already looks much better! Maybe the seperator line from NotesListView between the categories? I don't know because we don't have it in the drawer either...

I'd say fix the font color in dark mode and let's merge then (we always can improve later) :-)

dan0xii:
I've added another constructor to NavigationAdapter to enable/disable item highlighting to let the config activity just display all options in the dark theme textcolor.

screenshot_1524475336

> So the widget itself will not be able to be displayed in dark theme (yet)?

That's correct, I'll work on it next (in a separate PR :)). It'll need separate layouts for each widget and it will also default to whichever option the user has set in the main app as @korelstar suggested.

Issue #310, opened 2017-12-23
Confirm before deletion


enoch85:
A user just called me and was very frustrated because she accentually swiped left and removed a note containing ALL here passwords. The strange thing is that it didn't end up in the trashbin so I couldn't even restore it. Luckily I had a backup and we could revert the note from there, but still, I think it's WAY to easy to delete a note, as you don't even get a confirmation.

So, I suggest that you should add a confirmation when removing a note. Something like a popup asking "Are you sure?" or similar.

Please add this, it's important that notes don't get accentually deleted. Thank you!

cc @stefan-niedermann

stefan-niedermann:
Hi,

> The strange thing is that it didn't end up in the trashbin so I couldn't even restore it.

This is definitely a strange behavior. Can you reproduce this? A deleted note should be restoreable from trashbin.

I think, confirm-dialogs are a bad experience, since users are trained to click "ok"-buttons at each dialog. The app provides an Undo-Toast, why didn't the user use it? Didn't it work or was too little time to undo before the toast disappeared?

enoch85:
> why didn't the user use it?

The user is over 60 years old, and didn't notice the undo-toast. It should be user friendly for all ages, not just young people and tech-savvy folks.

Sorry, I can't reproduce it.

stefan-niedermann:
Okay, i have a strong opinion here, but let's get some further input! :)

@nextcloud/designers **Beside the problem, that the note did not land in the trashbin, as it definitely should have**: Please Up-Vote for Undo, Down-Vote for confirm-dialog

enoch85:
@stefan-niedermann Both would be nice.

In this specific case the note contained all her passwords, and I got a call late at night with someone more or less screaming "where are my passwords??!?!?!" You can blame the end user, that's the easiest, but in this case we/you could actually do something about it. :)

korelstar:
@enoch85 Did you figured out, why the note wasn't in the trash bin? Is the nextcloud app "Deleted files" activated?

I still like the idea of integrating the server side trash bin into the Android app.

enoch85:
@korelstar I have no clue. The trashbin was the first place I checked, but nothing there. 

Also, this was some time ago so I don't have any logs.

dan0xii:
[...]

PR #387, opened 2018-04-21
swipe to favorite

korelstar:
This branch was created by @stefan-niedermann.

I like the approach. After testing, I did a very small change and I think it's in a good state, now. However, there is still a discussion in #310, if this approach is sufficient. But with this PR, it's easier to track any changes on this.


stefan-niedermann:
@korelstar you might have read the commit commentar - do you have a good approach to change the background color based on the swipe-direction?

korelstar:
No, I didn't read the comment at that point. But after your hint I provided a solution for the colors (see new commit in this branch). The type of yellow is not perfect, so please feel free to chose a better one (`colors.xml`).

stefan-niedermann:
@jancborchardt can you help out with the color?

Background info: When swiping in the list from left to right, the background shall be yellow and as action the note is starred/unstarred instead of deleted.

When swiping from right to left the note gets still deleted (red background) - now we need an nice yellow color for starring/unstarring.

jancborchardt:
The icons and colors are:
- Favorite: #ffcc00 https://github.com/nextcloud/server/blob/master/core/img/actions/starred.svg
- Delete: #d40000 https://github.com/nextcloud/server/blob/master/core/img/actions/delete-hover.svg

The icons should be white, which you can achieve by simply opening them up in a text editor and chaning the fill attribute to `fill="#fff"`. :)

PR #396, opened 2018-05-07
Initial support for weblogin

tobiasKaminsky:
This enables weblogin for Nextcloud server.

 

Open questions
- [ ] Currently one has to enter http/https otherwise webview will fail
- [ ] on NC files app we have support for self signed cert, I added the code, but it is commented

If you need further help, please ping me :+1:


stefan-niedermann:
Great, thank you very much. Reason this PR is still open: I am struggeling with the self signed cert-stuff. I always use Let's Encrypt certs (which work well), but many people out there ise self signed certs. Therefore this has to work. Unfortunately i habe not very much experience in handling custom certs.

I think @schaarsc implemented the initial support: It would be great if @schaarsc could review this PR and give a "go" :)

tobiasKaminsky:
Thanks for the update, Stefan.
On our Files app, we have a warning if a self-signed cert is used, the user then has to manually accept it.
On weblogin flow we check it then with 
https://github.com/nextcloud/android/blob/master/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java#L403-L416

stefan-niedermann:
What's the reason you commented it in your PR?

tobiasKaminsky:
`NetworkUtils.isCertInKnownServersStore(cert, getApplicationContext()` is only available in android-library.

When manually accepting the cert, it gets stored via `NetworkUtils.addCertToKnownServersStore(m509Certificate, activity);`, so that we can check it later.

We can copy&paste those functions.
I am not sure if and how you manage self-signed certs, so I commented this out.

schaarsc:
I've not been following the development for quite some time, please give me a couple of days to catch up.
As I understand it, File-App is using NetworkUtils and Notes-App is using cert4android, but let me have a closer look...

stefan-niedermann:
Thank you very much @schaarsc, take your time.

And maybe you can contact me via email at info`@`niedermann.it please?

stefan-niedermann:
@korelstar you mean something like this: https://github.com/nextcloud/Android-SingleSignOn ?

Or did you mean the SelfSigned Cert stuff?

korelstar:
Oh, I wasn't aware of that. Does that library provide the same functionality like in this PR?

My intension was to provide both: the login fragment and ideally also the certificate handling. Then, all apps would have the same behavior for those problematic special configurations in our issues and all projects could reference the library project in this case.

korelstar:
@tobiasKaminsky 
Thanks, that sounds great! (for the others: see also nextcloud/Android-SingleSignOn#13)

Is there an estimated date when the library is "ready" (i.e. stable to use)? I.e., does it makes sense to wait for the library or does it worth to finish this PR?

David-Development:
I think there is no "stable" or "ready" state defined yet. I'm already using it in conjunction with the nextcloud news app on android and it's working stable for me right now. It would be great if other developers would start using / integrating it so that we can actually get more feedback, make it more stable and release this library into production soon.

eneiluj:
Thanks @tobiasKaminsky for this piece of code that I now use in PhoneTrack-Android.

Just in case someone wants to use it -> I had to make a little change that might be useful here too:

Execute the `URLValidatorAsyncTask` in `onTextChanged` for `field_url`. Otherwise the submit button is never enabled.

szaimen:
Any update on this? 
I think that's a really important feature for the app.
Thank you all for your work!

tobiasKaminsky:
Ready for final test and merge @stefan-niedermann 
(maybe also squash it…)

stefan-niedermann:
Finally in master!!!!! Thank you very much @tobiasKaminsky for your help on the Hackweek :) :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket:

Issue #353, opened 2018-03-30
Data loss when a note is open and device is offline during changes of other devices


chaosjug:
Steps to reproduce:
1. open a note in the app
2. turn of all network access (wifi etc.)
3. edit note on another device and make sure it is synced
4. turn on network access
5. leave the note --> app shortly shows as small sign, which means not synced yet if I read #344 correctly
6. old version is synced back --> changes made in 3. are lost

korelstar:
That's a known problem. However, solving this requires some changes on the server side and much work on the client side. I've written down some ideas on how to solve this in nextcloud/notes#56 and #161.

closing, because it's duplicate of #161

chaosjug:
This is not the same bug as #161. There you override changes from others with changes from the app. In my case I never edit the note at all. I don't even enter the editing mode.

But something changed in one of the last updates. When I wrote the initial bug report, I could trigger the bug every time. In version 0.16 (from fdroid) I tried it again and most of the time it correctly synced the changes. Only once the changes were overridden but I don't know what I did differently. I'll try to find a way to reproduce the bug and report back if I find something.

korelstar:
Okay, then the app "thinks" that the note was changed locally (the sync icon indicates this), even it wasn't. This would be indeed a new bug. It would be nice, if you could find a special property of the affected notes.

One idea: was the note written on Windows with Windows-style line-breaks? Due to problems in the Markdown-rendering, we convert Windows-style line-breaks to Unix-style. This could cause a local change.

Other idea: does this (only) happen for notes with URLs in it?

chaosjug:
I didn't have any problems in the last months, but we try to close notes now in the app, so there is less chance for this to happen.
I didn't check in detail, but there are two linux PCs and three Android phones connected to the same notes, so I wouldn't  expect any Windows-style line breaks. Nor do we use any special characters apart from what occurs in normal german texts.

stefan-niedermann:
Are you still able to reproduce this issue with version 2.0.0 (or greater)?

chaosjug:
I didn't notice any issuse since my last comment in February. Not sure since when I use a version greater 2.0.0 but I'm on 2.2.0 now.
Shall I close the bug?

stefan-niedermann:
Thank you for the feedback. Feel free to reopen in case you notice something again :)

PR #409, opened 2018-06-07
Fix save

### LINKED ISSUE #408: Save note when hiding application
When I edit a note and push the Home button quickly after a modification, the note is not saved and my changes are lost.
It would be great if the application could save the note when the app looses focus.

For example, the [ColorNote](https://play.google.com/store/apps/details?id=com.socialnmobile.dictapps.notepad.color.note) app does that (with a small "Saved" notification) and it's really convenient.

Thanks and keep up the great work!

korelstar:
- [x] fix save on pause/close/change fragment (fixes #408)
- [x] only save if something has changed (related to #353)

### DIFF
--- a/app/src/main/java/it/niedermann/owncloud/notes/android/activity/EditNoteActivity.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/android/activity/EditNoteActivity.java
@@ -148,7 +148,6 @@ public class EditNoteActivity extends AppCompatActivity implements BaseNoteFragm
 
     @Override
     public void onBackPressed() {
-        fragment.onPrepareClose();
         close();
     }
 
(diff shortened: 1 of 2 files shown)


korelstar:
Hmm. I don't think that it is my fault, that travis has failed:

```
FAILURE: Build failed with an exception.
* What went wrong:
Could not resolve all files for configuration ':app:debugCompileClasspath'.
> Could not find support-core-ui.aar (com.android.support:support-core-ui:27.1.1).
  Searched in the following locations:
      https://jcenter.bintray.com/com/android/support/support-core-ui/27.1.1/support-core-ui-27.1.1.aar
```

stefan-niedermann:
Yeah, it is not reachable in the browser either ...

grafik

Issue #379, opened 2018-04-13
Enhance Settings


stefan-niedermann:
Currently the most "nerdy" stuff in the settings gets the most attention / space: The self signed certificate stuff.

- [x] Rename "Account Settings" to "Account" since you know you already are in the settings
- [x] Display "guy"-icon at the left of "Account"
- [x] Display "eye"-icon at the left of "Display mode for Notes"
- [x] Merge the two certificate-related entries into on dialog. Call the Entry "Manage certificates" and put an appropriate icon at the left of it. (cc @schaarsc )

schaarsc:
@stefan-niedermann could you please explain this requirement
1. "merge cert entries into dialog"? something like this https://developer.android.com/guide/topics/ui/settings#Subscreens?
2. any preference for the icon? I'd suggest Material Design / https

stefan-niedermann:
1. Yes, exactly, a subscreen :)
2. I thought about this icon ("lock"): ![grafik](https://user-images.githubusercontent.com/4741199/41485959-d7f6c556-70e2-11e8-83e7-25e45293a5ec.png) Do you have any better idea?

PR #414, opened 2018-06-13
disable login until certificate is accepted.

schaarsc:
- submit button is disabled until url (incl. cert) is valid
- getX509CertificateFromError was copied from file app
- onReceivedSslError is ignored if certificate is known

PR #417, opened 2018-06-16
refactor settings

schaarsc:
PR for #379 
- move cert settings to subscreen
- add icons
- remove categories

### DIFF
--- a/app/src/main/res/drawable/ic_account_circle_blue_24dp.xml
+++ b/app/src/main/res/drawable/ic_account_circle_blue_24dp.xml
@@ -1,0 +1,9 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+        android:width="24dp"
+        android:height="24dp"
+        android:viewportWidth="24.0"
+        android:viewportHeight="24.0">
+    <path
+        android:fillColor="#0082C9"
+        android:pathData="M12,2C6.48,2 2,6.48 2,12s4.48,10 10,10 10,-4.48 10,-10S17.52,2 12,2zM12,5c1.66,0 3,1.34 3,3s-1.34,3 -3,3 -3,-1.34 -3,-3 1.34,-3 3,-3zM12,19.2c-2.5,0 -4.71,-1.28 -6,-3.22 0.03,-1.99 4,-3.08 6,-3.08 1.99,0 5.97,1.09 6,3.08 -1.29,1.94 -3.5,3.22 -6,3.22z"/>
+</vector>
--- a/app/src/main/res/drawable/ic_account_circle_grey_48dp.xml
+++ b/app/src/main/res/drawable/ic_account_circle_grey_48dp.xml
@@ -1,9 +1,15 @@
 <vector xmlns:android="http://schemas.android.com/apk/res/android"
-        android:width="24dp"
-        android:height="24dp"
+        android:width="48dp"
+        android:height="48dp"
         android:viewportWidth="24.0"
         android:viewportHeight="24.0">
+    <group
+        android:scaleX=".5"
+        android:scaleY=".5"
+        android:translateX="3"
+        android:translateY="6">
     <path
-        android:fillColor="#0082C9"
+        android:fillColor="#FF757575"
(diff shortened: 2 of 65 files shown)


stefan-niedermann:
awesome :) thank you! :rocket:

stefan-niedermann:
Nice work :)

- [ ] Make icons grey like in sidebar
- [ ] Is it possible that all list entries do have the same height whether or not they have a second line? (Account and certificates are currently smaller)
- [ ] Manage certificate does currently not show any top header bar, unlike account. It should have.
- [ ] The space between the icons and the text seems to be very big. is this some material standard or can we adjust it, so there is the same space like on the left side of the icons

device-2018-06-17-210853

stefan-niedermann:
Hmm... i think it is fine for now if we let the certificate stuff as it is. Maybe, after #396 we can discuss it again? I don't want another PR depending on #396 because we already have #414 and i don't know when we can finally merge #396.

korelstar:
That's fine.

schaarsc:
- I'll change the color to grey if you want, but we could also use black for notnight and grey for night ...?
- the height depends on the second line, I'd suggest to stick to the standard
- I'd suggest to keep certificate mgmt page as is, until #418 is decided.
- we could use bigger icons, like in Android-Settings/Account?

schaarsc:
image

Example to demonstrate some alternatives.
I'd choose bigger icon, notnight:blue, night:grey

Opinions?

stefan-niedermann:
cc @jancborchardt

jancborchardt:
The bigger icons seem ok – that also takes care of the whitespace being the same left and right.

But let’s use grey for both themes. Black on white is too extreme.

Very nice otherwise, good enhancement! :)

korelstar:
> the height depends on the second line, I'd suggest to stick to the standard

I agree, but I propose to add a second line with explanatory text for all items.

For example:

- Account: Connect to your Nextcloud instance
- Server certificates: Manage special settings for transport layer security (TLS)

Your welcome to improve my text examples. :smiley:

schaarsc:
notes

current impl with summary text. should I revert to only title?

stefan-niedermann:
Much, much better :)

@jancborchardt i think personally i like the smaller icons more, but that's up to you. also: While the second line is not necessairy i think it helps to scan all options quicker because all entries do have the equal height...

jancborchardt:
Looking at it now, I also think the smaller icons look better. It looks out of the ordinary on Android.

Equal lines don’t really help with scanning quicker if there’s more text to actually scan. ;) But more importantly: For display mode and theme, the second row actually shows relevant info – the current setting. With the additional text for account and certificates, the second row becomes irrelevant text which is probably not going to be read.

schaarsc:
so the final decision is:
- small icons
- not text for cert / account
???

jancborchardt:
@schaarsc yup! And, as @stefan-niedermann mentioned above:

> The space between the icons and the text seems to be very big. is this some material standard or can we adjust it, so there is the same space like on the left side of the icons

Then it would look much better.

schaarsc:
I don't know if the space can be changed, the example on material.io look similar to our settings with 24dp icons https://material.io/design/platform-guidance/android-settings.html#placement

notes

an alternative are bigger icons e.g. 36dp icons (https://github.com/stefan-niedermann/nextcloud-notes/pull/417#issuecomment-398842121 was 48dp)
notes-medium

stefan-niedermann:
Okay, let's go for the 24dp version for now :) looks great!

korelstar:
The only difference to the examples on material.io is the icon's alignment.

In your PR, the icons are not exactly below the arrow of the ActionBar's back button. Fixing this could improve the overall impression.

From material.io:
image

jancborchardt:
Yup – as @korelstar showed you only need to move them a bit to the right / center them.

If that’s done, we should go with the standard size icons – _not_ the bigger 36dp ones as they look strange.

schaarsc:
there is no option to control the icon position 
https://developer.android.com/reference/android/preference/Preference

it is possible to modify the icon and add hard coded offset
notes
but this may look different if something is changed on the underlaying Preference implementation.

korelstar:
Hmmm. No, that's dangerous. Maybe the icon size is not correct? How do other apps implement this?

I didn't found any app on my phone which has icons in it's preferences!

jancborchardt:
Let’s go for this as first step. :) We can always adjust later, but for now we’re blocked by details. ;)

stefan-niedermann:
I agree that modifying the hard icons with hardcoded offset is not a good idea. especially we can not estimate how this affects e. g. older android versions in combination with compat-libraries etc. etc. Though it is not perfect, we should go with the [version you posted here](https://github.com/stefan-niedermann/nextcloud-notes/pull/417#issuecomment-399847364) for now because it is way more stable and we do not have capacity to test each combination on every release:

version

@schaarsc can you please roll back to this version and resolve remaining conflicts? I will merge it then.

schaarsc:
patch provided in #433

Issue #422, opened 2018-06-18
Update rxmarkdown


korelstar:
There is a new release of RxMarkDown. Unfortunately, its API has changed. Now, there is also a variant of usage without RxJava. However, I don't know if that's advantageous.

We have to
- [ ] check which variant is better for us
- [ ] implement that variant

stefan-niedermann:
Yeap. There has also the possibility been added to set/unset checkboxes in view mode. It returns the linenumber and we would have to implement the logic ourself. i tried it on a local branch and i think this should be possible :)

PR #427, opened 2018-07-09
#422 Rxmarkdown update

stefan-niedermann:
This is a basic implementation of v0.1.2 of RxMarkdown using the Reactive X way.

It also implements toggling checkboxes from within the view mode.

problems left:

- [ ] We have to wait for v0.1.3 because of https://github.com/yydcdut/RxMarkdown/issues/50
- [ ] After toggling a checkbox the content is displayed as plain text and does not get processed anymore (unti switching to edit mode and back to view mode)
- [ ] toggling the last checkbox (last line) causes the last letter of this line to disappear. Not sure yet if this is a bug of RxMarkdown or a result of wrong usage.

I did not have any success implementing this without Reactive X (the checkbox-listener was not called - maybe a bug, too).

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -26,8 +26,10 @@ android {
 dependencies {
     implementation project(':cert4android')
 
-    implementation 'com.yydcdut:rxmarkdown:0.1.1-beta'
-
+    implementation 'io.reactivex:rxandroid:1.2.0'
+    implementation 'io.reactivex:rxjava:1.1.5'
+    implementation 'com.yydcdut:markdown-processor:0.1.2'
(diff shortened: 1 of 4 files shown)


stefan-niedermann:
v0.1.3 has been released https://github.com/yydcdut/RxMarkdown/releases/tag/v0.1.3

https://github.com/yydcdut/RxMarkdown/issues/50 should be fixed in this release. As soon as i have some time i will update and see how to fix the checkbox-issues.

stefan-niedermann:
Sorry to disappoint you, but i removed the support for toggling checkboxes for now because it still causes errors. 

I will open a new issue to track it since it basically **is possible now**. But to move on, we should first upgrade without adding more features.

Issue #428, opened 2018-07-11
widget's text appears white on white background


Keho:
With Nextcloud Notes v. 0.17.0 and 0.18.0,
the text in widgets (notes list, single note) appears white on the white widget background.
uninstalled nextcloud notes and reinstalled it in version 0.16.1, text is black.

LineageOS 14.1 Android 7.1.2
Tried with the Trebuchet launcher v14 and v15.

stefan-niedermann:
cc @dan0xii

Keho:
Same issue on the same phone (redmi 3s) with KISS launcher.
No issue on a mi mix 2 with Miui 9.1 Android 7.1.1, default launcher and KISS launcher.

dan0xii:
Does anyone else sometimes lose the title on the note list widget? I am seeing this happen on one of my devices and can't replicate it on anything else.

Keho:
It doesn't ring a bell but I'll pay attention.

Keho:
The « Notes » title is missing on a fresh install of v0.18.0
but not with 0.16.1 and 0.17.0.

TheManFromDelmote:
Redmi Note 4X, Lineage 15.1 Extended, build age 2 days ago. 
Widget has white text on white background on Evie launcher with both versions 0.18.0 & 0.17.0

Version 0.16.1 works fine. 
Thanks for the heads up Keho!!

dan0xii:
These remaining issues should be fixed in master.

stefan-niedermann:
@dan0xii is this completely fixed in master or only the workaround to hard coded use of the light mode?

dan0xii:
No, I just meant fixed as far as the end user is concerned. I should have a permanent fix coming soon.

PR #432, opened 2018-07-13
Fixes 428

dan0xii:
A quick fix for #428. The widgets appear to use the dark theme colours randomly and I need to spend more time figuring it out.


stefan-niedermann:
Thank you, i will let #428 open until a fix replaces this workaround.

PR #436, opened 2018-07-17
Added a reminder for notes

levnikmyskin:
Added a way to set reminders on any note.  
Reminders should be added in the local database in order to allow their deletion and recreation at boot times, however I'd wait for this pull request to be discussed before doing that.
I think some improvements could also be made on the dialog XML design.

About synchronization, I don't think Nextcloud notes allow for reminders at the moment, so this would be a local feature only.

Just let me know what you think about it :D


stefan-niedermann:
Hi @levnikmyskin and  @alessio,

thank you very much for your contribution.

Maybe the Tasks-app is a more appropriate tool to use for reminders? I am against adding this functionality to the notes app especially since feature comparison to the server side app is an important goal. Otherwise we would implement many features that are only supported locally and synchronisation with other tools like iOS clients or browser clients would not make sense. And the whole story is about synchronisation because otherwise one would not need a Nextcloud backend att all for some notes :)

I hope you can understand my concerns and i would be very happy if you are willing to contribute to other open issues in the issue tracker.

Regards

levnikmyskin:
Hi @stefan-niedermann,
thank you for your reply! 
What you say is totally reasonable, that's why I waited to dig deeper into the code and implement it thoroughly. I just find it really handy to set reminders on notes and wanted to give it a go. On the other hand, though, you made me discover the Tasks app.
I'll be glad to contribute on other issues on my free time :D 
Regards

PR #445, opened 2018-08-23
Introduce yellow favorite stars instead of outline

stefan-niedermann:
like @jancborchardt and @mario suggested:

- Instead of full grey stars use full yellow stars
- Instead of border grey stars use full grey stars

The changes apply to the widgets and the NotesListView but *not* to the Edit/View-Mode of a single note and *not* to the sidebar, where simple mono color icons look prettier imho.

device-2018-08-23-095707


mario:
Where are you?

stefan-niedermann:
@mario next to the fridge, what do you think? :laughing:

Issue #398, opened 2018-05-13
SSL issue at first Start


ei8ht187:
Hello
just installed the app through the Play Store. At start, during entering the server address (far from complete) my credentials i receive this message:

![screenshot_20180513-090309](https://user-images.githubusercontent.com/32712304/39964816-2b2030d4-568d-11e8-895d-d1cb829abdbf.png)

Therefore I don't trust the app actual. 
Device: Nokia 8
Android 8.1 
Patchlevel April 2018

self hosted Nextcloud 13.0.1 on nginx, TLS1.2, Let's encrypt cert

Can you please check this issue?

Regards

ei8ht

stefan-niedermann:
Tgis is strange since Let's Encrypt Certificates are known to work without problems. Especially the self-signed status should not ne shown.

Did you create your Let's Encrypt certificate recently or some time ago? Maybe the registration and validation process has not fully completed yet?

ei8ht187:
This message appears in the middle of my domain, the app can't know it. It appeared after https://cloud.star.....
But my domain has 7more characters before the country code .ch follows. Therefore I don't think it has to do with my Let's Encrypt cert.

ei8ht187:
Just checked with the F-Droid version: same problem, but strange is that i have just entered https, nothing more. Do you check for an already installed and configured Nextcloud app?
I will check my server later today. At least the F-Droid version shows the correct CN. Maybe i have to reengineer my serversetup...

korelstar:
The problem is, that the login activity tries to connect to the entered URL on every change of the input field, i.e. on every typed character. Therefore, the app tries to connect to `https://nextcloud.com` even if you are entering `https://nextcloud.comics.de/` (just an example).

The motivation for this check is to show a colored checkmark which indicates if there is a valid Nextcloud installation behind the current URL.

This behaviour of the app leads also to other problems, see #395. And beside the above mentioned problem with certificates, this is also a problem of privacy (the app contacts a wrong server).

I think it's time to overthink this and maybe (re)move the check.

stefan-niedermann:
Okay, here is my proposal: #446

The idea is to only make the URL check if the user blurs out of the field (and continues typing the username). This should prevent SSL, privacy and security issues caused by contacting wrong hosts.
What do you think?

Please review :)

stefan-niedermann:
Okay, i closed the PR and pushed it manually to master, sorry, i messed up the branches.
The issue should be fixed now.

PR #446, opened 2018-08-23
#398 Check the URL (and draw the check) only on blur

stefan-niedermann:
Checking the URL on every text change can lead to privacy issues (e.g. if a subdomain is also an existing domain and a request is done to this domain).


stefan-niedermann:
Sorry, messed up the branches. Don't care about it :joy:

Issue #359, opened 2018-04-10
Use monospace font


DriverXX:
Is it possible to add a configuration option to select the font used by the notes?
I access my notes also from PC and when I read them from my phone the layout is messed up, monospace font would solve that issue.

stefan-niedermann:
Hi @DriverXX,

i know there are beautiful mono space fonts out there and most markdown editors use mono space (at least in the edit mode). 

1. Do not at all (web interface does it neither)
2. Switch completely
3. With an option in the "View settings" area

I am struggeling with 3. because if we add it, there will be every day 3 feature requests ala "please add my font xyz-comic-sans-...". Providing all the available system fonts seems to be an overkill.

So we could provide "Sans serif, serif, mono space" without explaining which font is going to be used at all?

cc @nextcloud/designers for thoughts about mono-space font in the android notes app.

jancborchardt:
The problem is that we use Open Sans, where there isn't a Monospace equivalent (we don't need Serif). And that would look off.

korelstar:
A monospace font fits perfectly to a "source" mode, like in the Nextcloud-App "Markdown editor". I like this one and would be happy to see it in the Notes app (Nextcloud-App and Android-App), see also #246.

I think the comment from @jancborchardt applies to the web app, only. On Android, apps seem to be able to use a monospace font from the system.

jancborchardt:
The issue is that a monospace font just looks very technical, and it’s not easy to read. I can see though how some people might like it, so maybe a setting here is ok. Only to switch between »Normal« (meaning sans serif) and »Monospace« though. We don’t need a serif font.

stefan-niedermann:
Have a look at #447 for a new preference for monospace (as @korelstar mentioned, Android simple uses a system font with mono-space characters).

There are also screenshots in the PR @jancborchardt

PR #447, opened 2018-08-23
#359 Use monospace font

stefan-niedermann:
Introduces a new setting and displays edit and view mode in monospace if it is set.

device-2018-08-23-150032
device-2018-08-23-150001
device-2018-08-23-150016


stefan-niedermann:
Ah, i think i get now the issue. But on the other hand: Is this really what we want? I think we want to provide a WYSIWYG-editor which renders as soon as you type. If we divide again in an source-edit-mode and a rendering-mode we will go the same way as so many other editors out there: a split-pane-mode. In an ideal world i would expect that we could get rid of the "view-mode" completely and only rely on an Edit-AND-View-mode. I think this would be a step back.

DriverXX:
I think this is great, and it's enough for me, my request was simply to be able to use a monospace font.
That being said if you want really preserve the layout maybe you can consider to add a third "plain" view mode in which you do not use markdown.

jancborchardt:
The setting should be called (with second detail line):

> Monospace font
> Used in editing view

And if enabled, the monospace font should be used only in editing view of course – not in the preview. :)

This then won’t change anything about the default experience, where the edit mode is the default and looks nice by default. This setting basically just fixes it for people where there’s a mess in edit mode.

DriverXX:
> And if enabled, the monospace font should be used only in editing view of course – not in the preview. :)

I strongly disagree with this, why should we want a font change while changing view mode?
If I choose to use a monospace font I want it everywhere, I don't get your point on why we should have 2 different font for 2 different view mode

jancborchardt:
@DriverXX ok, let’s do the monospace font for both views then. :) But by default the option is off.

Issue #248, opened 2017-09-01
Refresh from server when in view mode


cstackpole:
Greetings,
Creating request based on #222 feedback.

When working on a group project, my team was editing files on the server and I was on my tablet getting the changes. Backing out of the file to refresh then going back into the file to view changes was a bit cumbersome. It is likely I will be working on a similar project in a similar fashion in the future.

It would be great if there was a way when in view mode, I could easily refresh that file from the server. It doesn't bother me if it is a button or a swipe down or something else. 

Thank you! Love this project and I'm getting tons of use out of it. :-)

stefan-niedermann:
Great idea. I will go for pull2refresh since it is common android pattern and we already use it in the app.

PR #450, opened 2018-08-24
WIP: #248 Refresh from server when in view mode

stefan-niedermann:
This PR basically works **but** whenever i set the new Content, it gets display but not rendered.

@korelstar any idea what might cause this behaviour?

Additionally: Currently this synchronizes all notes. Performance could be improved greatly if we only synchronize the current note.


korelstar:
> This PR basically works but whenever i set the new Content, it gets display but not rendered.
>
> @korelstar any idea what might cause this behaviour?

This may be related to the problem in #427:

> After toggling a checkbox the content is displayed as plain text and does not get processed anymore

I have no idea how to fix this. Maybe you should report this to upstream (Rxmarkdown).

PS. I will try to do the requested reviews in the next days, but I can't promise ...

stefan-niedermann:
https://github.com/yydcdut/RxMarkdown/issues/55

stefan-niedermann:
Some updates here: It seems that if one does not use `JavaRx` but [the normal implementation](https://github.com/yydcdut/RxMarkdown/#non-rx-usage), that the rendering after setting new content works correctly.

stefan-niedermann:
Aah, and yes. Toggling checkboxes seems to be possible with the normal implementation, too. It is currently active on this branch and i am happy if someone could help testing it.

@hanzei @rainerzufall @Offerel @jpggithub

PR #454, opened 2018-08-24
Fixes for the note list widget

dan0xii:
Fixes in the note list widget for a) alignment of the header app icon, b) alignment of the favourite stars next to the entries and c) the disappearing header text.


stefan-niedermann:
@dan0xii can you please post screenshots before/after for easier review?

Issue #360, opened 2018-04-10
Drawer header should show account name


stefan-niedermann:
The header in the drawer is currently quiet "dump" - it only shows the logo and the app name.

We should bump it up and show the name of the account and maybe the profile image. Let's see what's possible with the Nextcloud API ;-)

Clicking on the name or the profile picture should lead to the server settings.

PR #458, opened 2018-08-26
#360 Drawer header should show account name

stefan-niedermann:
device-2018-08-26-115546


jancborchardt:
@stefan-niedermann until there’s multi-account support (and as long as there’s only a single account configured in the app), a compact header even without showing the username + domain makes most sense.

stefan-niedermann:
okay, i will revert the layout to the compact header.

Issue #488, opened 2018-10-02
Dark mode does no longer work in widgets


stefan-niedermann:
Black font on dark background. cc @dan0xii 

Fllow-up to #482, #485 and #487 

![screenshot_20181002-085509](https://user-images.githubusercontent.com/4741199/46333949-907b7d80-c622-11e8-8efd-8a770717ce5b.png)

Keho:
I'm experiencing this with v0.22.0

dan0xii:
When do you see this? As soon as you place the widget? I'm having trouble reproducing it.

Keho:
Yes,
I found out how to reproduce the issue, it's a mismatch between the app theme setting and lineageos' theme setting.
(currently with 0.22.2)
1. if Nextcloud Notes is in Dark Mode turn it to White Mode, (current widgets stay dark (not sure if it's a side-issue)
2. if Lineageos' style is White turn it to Dark (settings, display, (advanced), style, style)
3. add a widget, the font is black on dark background.
This issue can happen when using lineageos' automatically changing style based on background or time.

What is weird though is that i'm not a user of lineageos's dark style (the option for me was set to automatic based on background, I think it's the default option, but everything was white), neither a user of Nextcloud notes' dark mode but somehow I experienced this issue (I think the LiveDisplay feature is messing with it too, I deactivated it during those tests).

dan0xii:
I think what's happening is that although the appwidget layouts can't change dynamically when daynight mode is set or unset, when they are re-inflated periodically android uses the colours from values-night if they are defined.

I've set the appwidget layouts to only use colours from values/colors.xml and have tested it with no problems. I don't have the theme option in lineageos (I'm on 7.x) so I can't test that but would appreciate it if you would @Keho. The code is in branch [nlwdt](https://github.com/stefan-niedermann/nextcloud-notes/tree/nlwdt).

Keho:
Sure, I'll have more time next week to test it.

Keho:
Alright, I installed it,
It seems to work, I can't reproduce the issue by using the steps mentioned above.
I'm not sure whether I built the app wrongly or not, but I experience lots of glitches, here is a list of what I experienced, I uninstalled and reinstalled the app multiple times, and restarted the phone many times too: 
* widgets could not be shown because of an error loading them.
* clicking on the list widget's tiltle « Notes » didn't open the app (it did nothing)
[...]

PR #503, opened 2018-11-26
Issue #488. Appwidget colours are not consistent in dark theme'd layout

### LINKED ISSUE #501: In dark mode the list bullet are black and not visible
The screenshot says it all (I think):

image

The bullet color should be the same as the font color

Version: 0.22.3

dan0xii:
Fixes #501.


stefan-niedermann:
hm. this does seem to not fix the issue (while not making it worse, therefore i will let this in master branch) :-/

PR #517, opened 2019-01-19
Markdown util needs to hardcode unordered list colour.

dan0xii:
getMarkdownConfiguration will set unordered list colour depending on whether or not dark theme is enabled.


dan0xii:
This is a follow-up to #488. As it stands, if you place the single note widget on the home screen and then change the theme in the app, the snw's markdown text will change colour dynamically whereas the rest of the layout is static.

Before:
screenshot_1547927491

After:
screenshot_1547927609

dan0xii:
I couldn't do that because when the theme changes in the main app the text colour wouldn't change dynamically the way it does now. Hope that makes sense !

Issue #276, opened 2017-10-15
Font size


eemantsal:
Hi.

I have read [this issue](https://github.com/stefan-niedermann/nextcloud-notes/issues/189), but I want to insist and ask the developer to reconsider the matter, not because line breaks that make the notes look more or less pleasant, but for a very simple reason: seems that mature people have difficulties to read, but especially to edit their notes. I've verified with two persons I have "evangelized" to abandon Google and come to the "free world".
They don't have any visual defficiency beyond what's normal in the sixties and seventies: presbyopia, something that practically all of us shall suffer when we turn 40 - 45 y. o. and beyond. While their mobiles'  UI font size seems sufficient to read most of the apps, it turns out that NC Notes app's summary view uses a much smaller font. But the major problem comes when editing the notes: the font size makes difficult to handle the cursor, make selections, to copy/paste, etc, even with their glasses.
I must add thay while I have no problems to see the text clearly, I'd thank if I could set the interface font size bigger because the size of my fingers. Not that they are enormous, I think are normal for a near to 1,90 m man, but sometimes it's hard to place the cursor where I want; it hapens also when typing an email or chatting, but in those kind of applications I don't use text format nor edit what I type so often, so I don't care much. But it's different in an editor like this notes editor.

So, the fixed font size is more than an aesthetic problem, it's really an usability one. I hope you may reconsider your position. I invite you to let use the app to your aunt, or your father, and ask them about their experience, I think it might be rather similar to my two "subjects".

Regards

korelstar:
Did you changed the device's font size? This can be done in the Android 
settings in the section "Display" -> "Font size" or in "Accessibility" 
-> "Font size". This has also effects to the Notes app's font size.

I think, it's more productive to change the font size of the whole 
device instead of changing the font size in every single app. Isn't this 
sufficient?

eemantsal:
I know, I know text size can be changed for all the Android UI, but the bigger the size you set, the less information fits on your screen. In my personal case I have no problem in reading the text in my phone's UI nor in the notes app's one, but I can't handle editions as well as my girlfriend, for example, who has smaller and thinner fingers. I don't need bigger fonts in a system-wide manner, not even for just reading my notes in the notes app, but would benefit from an option to increase it in the edition mode.
[...]

PR #575, opened 2019-08-06
Added preference "Font Size".

keeshii:
The default font size is quite big, reading longer notes was troublesome. I decided to add an additional setting in the preferences. Now user can change the font size.

### DIFF
--- a/app/src/main/java/it/niedermann/owncloud/notes/android/fragment/BaseNoteFragment.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/android/fragment/BaseNoteFragment.java
@@ -5,6 +5,7 @@ import android.app.Fragment;
 import android.app.FragmentManager;
 import android.app.PendingIntent;
 import android.content.Intent;
+import android.content.SharedPreferences;
 import android.content.pm.ShortcutInfo;
 import android.content.pm.ShortcutManager;
 import android.graphics.drawable.Icon;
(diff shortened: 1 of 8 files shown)


stefan-niedermann:
For reference: #276

The displayed font size depends on the system settings currently.

@jancborchardt do you think it is a needed option? It has been requested by 4 users until now.

stefan-niedermann:
@keeshii thank you for the contribution, code-wise it looks very good. I will have a look about how it feels. While i declined a seperate setting for the font size in the past, i will reconsider it as the focus of the Notes app is on the texting and it might make sense.

jancborchardt:
Right, the automatic font-size inheritance is a problem for accessibility here. So in this case a setting is fine. :)

@keeshii thank you for your contribution! Can you post a screenshot of the change for a quick design review?

keeshii:
It's standard `ListPreference`. I had hard time deciding, which icon should I use. Finally I have selected the icon: [format_size](https://material.io/resources/icons/?icon=format_size&style=baseline).

Screenshot_20190808-062700_Notes
Screenshot_20190808-062715_Notes

jancborchardt:
@keeshii nice! Icon and entry looks good. :+1:

Can you explain the sizes? I would have expected Small, Medium and Large, with "Medium" being the default.

keeshii:
Sure.

At the beginning I was considering using the three options you mentioned: Small, Medium and Large. Such values are defined in the [TextAppearance](https://stackoverflow.com/a/11590581) resource of the android sdk. TextAppearance is using following sizes:
- Small = 14sp
- Medium = 18sp
- Large = 22sp

The unit "sp" means "scalled pixels". They represent the number of pixels, but the value is scalled by the DPI of the screen. The "sp" values are also affected by the font size setting in the system preferences, which makes it the best choice for defining the font sizes in the application.

I was playing a bit with those values on my phone. After a while I decided to use following options:
- Tiny = 14sp
- Small = 16sp
- Medium = 18sp
- Large = 22sp

14sp is quite small. Such size is good for captions on small UI elements, but not much for texting. Maybe on tables this size would be ok. I decided to leave this option as "Tiny".

Medium (18sp) is actually the system default. This size application was using before my MR. In my opinion it is big, so I put also something in between - 16sp.

... and Large is really Large. Good for old people.

jancborchardt:
> 14sp is quite small. Such size is good for captions on small UI elements, but not much for texting.

Then I would just leave the "Tiny" option out. :)

3 options of Small, Medium and Large are very quick to understand, and "Medium" is going to be a good default. As soon as there’s another option, you don’t know anymore which is the inherent default and also as you said it’s quite small anyway.

What do you think @stefan-niedermann?

stefan-niedermann:
Yes, i agree :)  @keeshii can you adjust the PR? :)

keeshii:
> Yes, i agree :) @keeshii can you adjust the PR? :)

No problem. Done.

stefan-niedermann:
Great, thank you for your contribution :) It might take some time until this will be released on Play Store because we usually are waiting for some translations via transifex (which might take some time).

jancborchardt:
@keeshii welcome to Nextcloud Notes! :) More contributions are always very welcome!

PR #585, opened 2019-09-10
fix background for code in dark mode

korelstar:
The background-color for code elements was the same in dark mode and light mode. Therefore, in dark mode, code elements are not easy to read. With this PR, the chosen colors do respect the current theme and use a much darker background color in dark mode.

## Before
Screenshot

## After
Screenshot

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -39,7 +39,7 @@ dependencies {
     implementation 'com.jakewharton:butterknife:10.1.0'
     annotationProcessor 'com.jakewharton:butterknife-compiler:10.1.0'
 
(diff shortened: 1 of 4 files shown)


stefan-niedermann:
Does the removal of the Compat-usage has any negative drawbacks, especially for old android versions?

korelstar:
No, but I missed that this `getColor` is deprecated now. I will provide an update in the evening.

korelstar:
fixed; ready for review, now! :smiley:

stefan-niedermann:
thanks :)

PR #58, opened 2016-01-21
Merge pull request #1 from stefan-niedermann/master

HeaDBanGer84:
pull RecyclerView branch

PR #586, opened 2019-09-12
fix search in category

korelstar:
The search term wasn't applied in all cases, when you search inside of a category.

I will have to test this further.


stefan-niedermann:
Is this related to #58) ?

PR #598, opened 2019-10-04
improve timeslots in notes list

### LINKED ISSUE #539: Date divider in list view
When I started using the app I didn't even realize what it was meant for. I'm not sure if it's got any use for somebody.

Maybe the date could be shown for each note with a mix of relative times for recent notes and absolute times like when note older than a week or so. 

https://cloud.addictivetips.com/wp-content/uploads/2017/07/pin-notes-ios-11.jpg

korelstar:
Use the same approach as in https://github.com/nextcloud/notes/pull/313.

Fixes #539.


stefan-niedermann:
Thank you very much :) 

If you are interested: have a look at `multi-account` branch, where i started a implementation based on Single Sign on supporting multiple accounts and completely getting rid of self signed certificate handling. Attention: It's still very basic and work in progress.

korelstar:
Nice :-)

I will have a look if I find some time ...

PR #610, opened 2019-10-24
[WIP] SSO: return response headers

tobiasKaminsky:


stefan-niedermann:
So i guess the solution you chose is wrapping the inputstream into a response object which also contains the headers?

Thank you very much @tobiasKaminsky !!

cc @korelstar

tobiasKaminsky:
Yes.

Please have a look at the SimpleDateFormat stuff.
For some reason it seems to not parse the date correctly.

PR #619, opened 2019-12-15
Skip dropping automatic indices

Unpublished:
They cannot be dropped manually:
`index associated with UNIQUE or PRIMARY KEY constraint cannot be dropped`

> For automatically created indices (used to implement the PRIMARY KEY or UNIQUE constraints) the sql field is NULL.

From https://sqlite.org/faq.html#q7

Can be reproduced by just upgrading to 2.0.0 (without login).

While at it I used try-with-resources to automatically close the cursor.

### DIFF
--- a/app/src/main/java/it/niedermann/owncloud/notes/util/DatabaseIndexUtil.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/util/DatabaseIndexUtil.java
@@ -27,11 +27,14 @@ public class DatabaseIndexUtil {
     }
 
     public static void dropIndexes(@NonNull SQLiteDatabase db) {
-        Cursor c = db.query("sqlite_master", new String[]{"name"}, "type=?", new String[]{"index"}, null, null, null);
-        while (c.moveToNext()) {
-            Log.v(TAG, "Deleting database index: DROP INDEX " + c.getString(0));
-            db.execSQL("DROP INDEX " + c.getString(0));
+        try (Cursor c = db.query("sqlite_master", new String[]{"name", "sql"}, "type=?", new String[]{"index"}, null, null, null)) {
+            while (c.moveToNext()) {
+                // Skip automatic indexes which we can't drop manually
+                if (c.getString(1) != null) {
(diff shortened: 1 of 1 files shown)


stefan-niedermann:
Thank you for the contribution, but maybe you can clearify, what exactly the problem is? I can not understand it.

Upgrading from `1.0.2` to `2.0.0` works without any noticable problem for me. Is there some kind of exception in the Logcat for you? Or does the app crash on upgrade?

Unpublished:
> Upgrading from `1.0.2` to `2.0.0` works without any noticable problem for me. Is there some kind of exception in the Logcat for you? Or does the app crash on upgrade?

Sorry for not being clear.
It kept crashing for me on startup after upgrade.
I know it contained the mentioned `index associated with UNIQUE or PRIMARY KEY constraint cannot be dropped` error with something like `sqlite_autoindex_accounts1` and it was caused by the edited method.

But I don't have the logcat anymore...
I'll try to reproduce and post the complete logcat.

stefan-niedermann:
@Unpublished, since you seem to know little about the SQLite stuff: I tried to add an foreign key from the Notes table (account-id) to the accounts-table (id) in version 2.0.0.

Unfortunately i am not sure if i did this right, because in my SQLite browser i can not see any foreign key association.

Maybe you want to have a look at this or have a hint for me? :rocket:

Unpublished:
First, thanks for merging and I had to look this up as well :)


I couldn't find it in SQLite browser as well.
Then I opened the db in IntelliJ (there is a database plugin) and there the foreign key is displayed.
Afterwards I discovered it in SQLite browser via selecting `NOTES` in `Database Structure` tab, `Modify Table`, scroll to right in the new popup and there it is under `Foreign Key`.

stefan-niedermann:
Thank you for having a look! Well well, know your tools, shame on me :laughing:

Would be great to see you again contributing in the future :)

Unpublished:
You never stop learning :)

Sure, I'll keep contributing from time to time to the whole nextcloud ecosystem. Might be hard for the notes app as I'm quite happy with it ;)

Issue #604, opened 2019-10-12
Crash upon cutting the last item of a bullet list


vchryssos:
Notes version: 1.0.0
Android version: 9.0

Reproduceability:
- make a list of items, using "*" notation for each list item
- select last item
- cut
- *Notes* crashes with following log

Log:
Version: 1.0.0

```
java.lang.IndexOutOfBoundsException: charAt: 469 >= length 469
	at android.text.SpannableStringBuilder.charAt(SpannableStringBuilder.java:123)
	at android.text.CharSequenceCharacterIterator.current(CharSequenceCharacterIterator.java:58)
	at android.text.CharSequenceCharacterIterator.setIndex(CharSequenceCharacterIterator.java:83)
	at android.icu.text.RuleBasedBreakIterator.CISetIndex32(RuleBasedBreakIterator.java:1126)
	at android.icu.text.RuleBasedBreakIterator.isBoundary(RuleBasedBreakIterator.java:503)
	at android.text.method.WordIterator.isBoundary(WordIterator.java:95)
	at android.widget.Editor$SelectionHandleView.positionAtCursorOffset(Editor.java:5467)
	at android.widget.Editor$HandleView.invalidate(Editor.java:4577)
	at android.widget.Editor$SelectionModifierCursorController.invalidateHandles(Editor.java:6145)
	at android.widget.Editor.invalidateHandlesAndActionMode(Editor.java:1985)
	at android.widget.TextView.spanChange(TextView.java:9926)
	at android.widget.TextView$ChangeWatcher.onSpanRemoved(TextView.java:12548)
	at android.text.SpannableStringBuilder.sendSpanRemoved(SpannableStringBuilder.java:1292)
	at android.text.SpannableStringBuilder.removeSpan(SpannableStringBuilder.java:500)
	at android.text.SpannableStringBuilder.removeSpan(SpannableStringBuilder.java:799)
	at android.text.SpannableStringBuilder.removeSpan(SpannableStringBuilder.java:787)
	at com.yydcdut.markdown.live.ListLive.updateListSpanBeginning(ListLive.java:641)
	at com.yydcdut.markdown.live.ListLive.onTextChanged(ListLive.java:149)
	at com.yydcdut.markdown.live.LivePrepare.onTextChanged(LivePrepare.java:81)
	at com.yydcdut.markdown.MarkdownEditText.onTextChanged4Controller(MarkdownEditText.java:252)
[...]

Issue #477, opened 2018-09-15
Crash when cutting title


Keho:
Hi,

1. Create a new note with the following text:
`# Test`
2. Select either everything or everything but the number sign (the space must be selected).
3. Press "Cut". The app should crash.

Nextcloud Notes 0.22
AnysoftKeyboard 1.9.2055
LineageOS 15.1 Android 8.1.0

Crashes on olders version of the app too (down to 0.14.2 at least, didn't try earlier than that).

stefan-niedermann:
Not reproducable by me. (pretty much the same setting).

Can you share the exact note content and the category and if it is marked as favorite, so i can set up the exact same environement?

Keho:
The exact note is just **# Test**
Tried with no category and different categories and it crashes each time.
It also crashes whether the note is marked as favorite or not.

stefan-niedermann:
I have been able to reproduce the issue on my physical device. I will try to investigate the behavior.

stefan-niedermann:
```
09-16 13:51:08.566 27494-27494/it.niedermann.owncloud.notes E/AndroidRuntime: FATAL EXCEPTION: main
    Process: it.niedermann.owncloud.notes, PID: 27494
    java.lang.IndexOutOfBoundsException: charAt: 2 >= length 2
        at android.text.SpannableStringBuilder.charAt(SpannableStringBuilder.java:123)
        at android.text.CharSequenceCharacterIterator.next(CharSequenceCharacterIterator.java:67)
        at android.icu.text.RuleBasedBreakIterator.handleNext(RuleBasedBreakIterator.java:1311)
        at android.icu.text.RuleBasedBreakIterator.next(RuleBasedBreakIterator.java:416)
        at android.icu.text.RuleBasedBreakIterator.rulesFollowing(RuleBasedBreakIterator.java:742)
        at android.icu.text.RuleBasedBreakIterator.following(RuleBasedBreakIterator.java:696)
        at android.icu.text.RuleBasedBreakIterator.isBoundary(RuleBasedBreakIterator.java:932)
        at android.text.method.WordIterator.isBoundary(WordIterator.java:95)
        at android.widget.Editor$SelectionHandleView.positionAtCursorOffset(Editor.java:5005)
        at android.widget.Editor$HandleView.invalidate(Editor.java:4298)
        at android.widget.Editor$SelectionModifierCursorController.invalidateHandles(Editor.java:5660)
[...]

Issue #549, opened 2019-04-04
Replace RxMarkdown with Markwon


quthla:
Some tasks left, but i think we are already on a good way

- [x] Syntax Highlighting (needs another lib i think)
- [x] Different Header levels do not work properly yet in edit mode
- [x] Fix coloring on search
- [ ] Check if we can prevent selection of tasks when toggling checkboxes
- [x] Check if a checkbox can be toggled when clicking on the checkbox directly (not the text)
- [ ] Check external links, especially fix or remove links in checkboxes (affects maybe only [edit mode](https://github.com/noties/Markwon/issues/199)?)
- [ ] Make sure that internal links work
- [ ] Styling (should be as close as possible to the webapp)
  - [x] Codeblocks should always use monospace font
  - [ ] Closing fence of cosing blocks are different from opening blocks
- [x] Extract plugins where possible to make them reusable
- [x] Centralize the configurations
  - [x] Respect monospace font setting
  - [x] Respect font size setting
- [ ] Use `Markwon` on all places where previously `RxMarkdown` has been used (e. g. widgets)

stefan-niedermann:
Looks interesting. Does it support inline-editing?

tnyeanderson:
Keep an eye on this, maybe we can get the same platform for mobile and the webapp:

https://github.com/nextcloud/notes/issues/204

tnyeanderson:
I don't think Markwon will work with EditText. [It's not built for it](https://github.com/noties/Markwon/issues/97#issuecomment-460990096). We do need to find a replacement though. This issue should be retitled and used to track finding a new markdown library.

tnyeanderson:
Here's one to look at: https://github.com/crescentflare/SimpleMarkdownParser

stefan-niedermann:
Unfortunately only supports

```
headers (#), text styles (italics and bold), strike through text, lists and links
```

which is way less then the current lib.

tnyeanderson:
Definitely, it's a newer library (says more support is coming.... who knows when). It's surprisingly difficult to find renderers that are actively maintained and allow inline editing :(

What else would need to be included? Those styles are supported by nextcloud/notes using SimpleMDE, and only supports the following in addition:
* Quotes
* Images
* Tables
* Code

We should also include **tasks** functionality (with clickable checkboxes)... but to get SimpleMarkdownParser up to par might not be as bad as we think...

What other syntax would be needed other than those listed?

[...]

PR #622, opened 2019-12-16
Workaround for #477 and #604

stefan-niedermann:
Unfortunately, we can not solve the problem 100% with the limited `RxMarkdown` library. This is the attempt to have a workaround which does not cause a complete app crash.

I had a look at the cause and the `EditText` seems to only "know" the first line at the moment when the "cut" operation is executed. The selection though is correct (and not in the first line). This seems to cause the `IndexOutOfBoundsException`.

This is the attempt to catch the exception and reset the selection to prevent the OutOfBounds. Unfortunately this covers only a few little cases but i can not offer much more at the current time.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -43,7 +43,7 @@ dependencies {
     implementation 'com.github.bumptech.glide:glide:4.10.0'
     annotationProcessor 'com.github.bumptech.glide:compiler:4.10.0'
 
-    implementation 'com.jakewharton:butterknife:10.2.0'
+    implementation 'com.jakewharton:butterknife:10.2.1'
     annotationProcessor 'com.jakewharton:butterknife-compiler:10.2.0'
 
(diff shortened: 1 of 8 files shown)


stefan-niedermann:
Maybe @vchryssos and @Keho can try to test if their special issues could be solved with this branch (i attached this branch as a `.apk`-test-file) which you can uncompress and install for testing purposes.

[app-debug.apk.zip](https://github.com/stefan-niedermann/nextcloud-notes/files/3967584/app-debug.apk.zip)

Sorry, but i can not offer much more until #549 is done.

vchryssos:
> Maybe @vchryssos and @Keho can try to test if their special issues could be solved with this branch (i attached this branch as a `.apk`-test-file) which you can uncompress and install for testing purposes.

I tested the .apk. Seems to work fine. The problem does not persist.
Great work @stefan-niedermann thanx for working this out!

Issue #528, opened 2019-02-09
Save the excerpt in a separate column


stefan-niedermann:
Save the excerpt in a separate column (with index) and then do not load the full note (i.e. load it only without the actual text), when only the list of notes is required. (Follow-up of #154).

Issue #154, opened 2016-12-10
Improve user experience on slower devices


korelstar:
I've tested the app on my old Samsung Galaxy Gio (from 2011) and found some bottlenecks: Especially,   loading the list of notes from the database and the synchronization are very slow.

## Loading the list of notes from the database
- [x] Introduce database index to speed up sorting
- [ ] Maybe it could be helpful to save the excerpt in a separate column (with index) and then do not load the full note (i.e. load it only without the actual text), when only the list of notes is required.

## Synchronization
- [x] Synchronization takes very much time, because always all notes are transferred and compared. This could be improved, but requires changes on the server site (e.g. check if the `ETag` of the notes file has changed).

stefan-niedermann:
@korelstar the synchronisation part has already been done, right? Maybe we than should drop this paragraph while the other proposals still make sense.

korelstar:
Yep, you're right. I've updated the text and introduced check-marks to track the progress.

stefan-niedermann:
Follow-Up in #528.

PR #641, opened 2019-12-27
#528 Save the excerpt in a separate column

stefan-niedermann:
Do you want to make a review since you requested this originally in https://github.com/stefan-niedermann/nextcloud-notes/issues/154

### DIFF
--- a/app/src/main/java/it/niedermann/owncloud/notes/persistence/NoteSQLiteOpenHelper.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/persistence/NoteSQLiteOpenHelper.java
@@ -266,7 +266,6 @@ public class NoteSQLiteOpenHelper extends SQLiteOpenHelper {
                 db.update(table_notes, values, key_id + " = ? ", new String[]{cursor.getString(0)});
             }
             cursor.close();
-            DatabaseIndexUtil.createIndex(db, table_notes, key_excerpt);
         }
     }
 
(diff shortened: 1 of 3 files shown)


korelstar:
Oh, nice! Yep, I will review and test this in detail, soon. Looks good on my first view :-)

app/src/main/java/it/niedermann/owncloud/notes/persistence/NoteSQLiteOpenHelper.java
```diff
@@ -239,6 +257,17 @@ public void onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
...
+        if (oldVersion < 10) {
+            db.execSQL("ALTER TABLE " + table_notes + " ADD COLUMN " + key_excerpt + " INTEGER NOT NULL DEFAULT ''");
+            Cursor cursor = db.query(table_notes, new String[]{key_id, key_content}, null, null, null, null, null, null);
+            while (cursor.moveToNext()) {
+                ContentValues values = new ContentValues();
+                values.put(key_excerpt, NoteUtil.generateNoteExcerpt(cursor.getString(1)));
+                db.update(table_notes, values, key_id + " = ? ", new String[]{cursor.getString(0)});
+            }
+            cursor.close();
+            DatabaseIndexUtil.createIndex(db, table_notes, key_excerpt);
```
korelstar: I don't think that we need an index for this column, since we don't search for values in it and we don't sort by it. What was the reason for you to add an index?

stefan-niedermann:
Nice :) I noticed a problem while upgrading from previous versions (and patched that in this branch) - maybe you could do a test from an old version too when you are over it.

app/src/main/java/it/niedermann/owncloud/notes/persistence/NoteSQLiteOpenHelper.java
stefan-niedermann: Agree, i think we can dump it.

stefan-niedermann:
Thank you for the review and the fixes :)

Issue #623, opened 2019-12-16
Links with relative URLs


lenzls:
Hello,

imagine a note like
```
 1. [link!](https://www.example.com)
 2. [link!](example.com)
 3. [link!](56366)
```

### Behaviour nextclound notes webapp:
1. is a link to `https://www.example.com`
2. Is a relative link like `https://<nextcloud>/apps/notes/note/example.com`
3. Is a relative link like `https://<nextcloud>/apps/notes/note/56366`, which is a link to another note and allows me to link different notes with each other

### Behaviour nextclound notes android:
1. is a link to `https://www.example.com`
2. Is not interpreted as a link at all
3. Is not interpreted as a link at all

-----

I think it would be nice to also support some kind of local linkage between notes. I'm currently trying out to use the notes as kind of a wiki-like system.

And while doing that I stumbled over the issue above.

stefan-niedermann:
Hi,

i am open to implement this, but only if the server app will implement this too (we have to male sure that we both use the same linking system / uri scheme).

@korelstar what do you think about it?

Just for reference some discussions and an old PR: https://github.com/stefan-niedermann/nextcloud-notes/pull/202

lenzls:
@stefan-niedermann 
That's nice to hear!

Forgive me my ignorance, but doesn't the webapp already implement this?

As shown in the issue, when using relative urls in the webapp, the linking already works. 
The link `[link!](56366)` is rendered as `https://<nextcloud>/apps/notes/note/56366`. And when clicking this link, my browser correctly takes me to the linked note. 

Of course the android app needs to behave a little different, since linked pages are normally rendered in a browser and not in the nextcloud-notes application.

stefan-niedermann:
Indeed. The id is interpreted as a relative URL part which causes a click on it to jump directly to that note. Well then :) I think, one could grap out #202 and work on this base :)

lenzls:
I think I might give it a try.

But before I'd like to get the behaviour straight.

### proposal
#

lenzls:
I think I might give it a try. 

But before I'd like to get the behaviour straight.

## proposal

Given the following markdown examples

 1. `[link!](https://www.example.com)`
 2. `[link!](example.com)`
 3. `[link!](<id-of-existing-note>)`
 4. `[link!](<string-that-looks-like-an-id-but-there-is-no-such-note>)`

in view mode these snippets will be rendered as

[...]

PR #646, opened 2019-12-28
Note links

lenzls:
Hey everybody,

so I have now a working version of the note-links proposal from #623 . But due to a bug in rxMarkdown things got a little hacky…

### Details

Links that do __not__ start with `/https?/` are ignored by rxMarkdown. It completely ignores constructs like `[link](example.com)` and therefore also `[link](356675)` (1). That also means that I can't hook into the onLinkClick callback to open a note-link in the app.

In order to work around this, I introduced the following workaround:

Before the markdown gets passed to rxMarkdown I traverse through and replace all links of the format `[<some-text>](<some-existing-note-file-id)` with 
`[<some-text>](https://nextcloudnotes/notes/<some-existing-note-file-id>)`. Basically I add a dummy prefix to the ids to create valid urls.

Then the markdown string gets passed to rxMarkdown as usual. Since these note-links now start with `https`, rxMarkdown correctly interprets them as links.

Now I can hook into the onLinkClick callback and on every link click I execute the logic:
 - Is the link of the structure `https://nextcloudnotes/notes/<someId>` _then_ open it in the nextcloud-notes application.
 - Is the link not of this structure _then_ open it in the standard webbrowser (behaviour as before).

-----
(1) Actually I think, that rxMarkdown is even more broken. As reported in [65](https://github.com/yydcdut/RxMarkdown/issues/65), the library is technically also ignoring links like `[<text>](https://example.com)`. But since they also implemented autolinking of urls in text (`<text> https://example.com <moretext>`), the links are recognized in the end. But not because they are in markdown link syntax, but because the string contains `http<something>`. 

Unfortunately I don't expect this to get fixed without forking the library (if I look at their activity)…

### Conclusion

I know this replacement-logic is really not nice, but I didn't see a better approach. As long as rxMarkdown does not handle markdown links correctly, we can't influence the on-click behaviour.

If you can think of something nicer, I'd be happy to dig deeper :)

While technically, the users shouldn't realize the workaround at all, but in reality there is a fraction of a second, after the text gets replaced where the replaced links are visible to the endusers. I guess we could somehow get rid of that if you decide that the PR in in principle mergeable.

But I would also understand if you don't like the current state.

Let's see what you think and hopefully one of you has a better idea.


lenzls:
By the way, I have experience in Java but not on Android. So forgive me if I did some things a little unconventional. 
I was surprised that it actually uses a different JDK and we are stuck with some subset of jdk8. Not even streams and stuff.

Also I couldn't get your existing tests to run and figured that they were broken in the beginning. If that assumption wasn't true, I can revert my changes there.
The old ones still don't run, but at least they compile now and the newly added ones from me run through :)

stefan-niedermann:
> By the way, I have experience in Java but not on Android. So forgive me if I did some things a little unconventional.

Well i'm a web developer actually, so nothing to shame for :laughing: 

> I was surprised that it actually uses a different JDK and we are stuck with some subset of jdk8. Not even streams and stuff.

Yeah :cry: at least lambdas are available.

Okay, basically i think we can go this way, even if it might not be optimal.

>  in reality there is a fraction of a second, after the text gets replaced where the replaced links are visible to the endusers

I will check this with a slow virtual device, maybe we can tune the performance here a little bit, so noone will notice it.

lenzls:
> Okay, basically i think we can go this way, even if it might not be optimal.

Nice, and while working on #549 we should be able to replace the hacky part easily anyway.

> I will check this with a slow virtual device, maybe we can tune the performance here a little bit, so noone will notice it.

If it helps, I had the issue on my actual phone. It is a Xiaomi A1 with Android 9.

app/src/main/java/it/niedermann/owncloud/notes/util/NoteLinksUtils.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class NoteLinksUtils {
+
+    @VisibleForTesting
+    static final String RELATIVE_LINK_WORKAROUND_PREFIX = "https://nextcloudnotes/notes/";
+
+    private static final String linksThatLookLikeNoteLinksRegEx = "\\[[^\\]]*\\]\\((\\d+)\\)";
```
stefan-niedermann: This regex looks very complicated. Wouldn't something like `^\[.+\]\(\d+\)$` be easier? To be honest, i do not understand your regex, there are so many escapes :eyes:

app/src/main/java/it/niedermann/owncloud/notes/android/fragment/NotePreviewFragment.java
```diff
@@ -113,6 +139,15 @@ public void onNext(CharSequence charSequence) {
...
 
+    private Set<String> getExistingNoteRemoteIds() {
+        List<DBNote> notes = db.getNotes(note.getAccountId());
+        Set<String> noteRemoteIds = new HashSet<>();
+        for (DBNote note : notes) {
+            noteRemoteIds.add(String.valueOf(note.getRemoteId()));
+        }
+        return noteRemoteIds;
+    }
+
```
stefan-niedermann: We should move this into NoteSQLiteOpenHelper class as a direct SQL-query for performance reasons.

app/src/main/java/it/niedermann/owncloud/notes/android/fragment/NotePreviewFragment.java
```diff
@@ -19,23 +21,32 @@
...
 public class NotePreviewFragment extends BaseNoteFragment {
-    
+
     private static final String TAG = NotePreviewFragment.class.getSimpleName();
 
     @BindView(R.id.single_note_content)
     RxMDTextView noteContent;
 
+    private NoteSQLiteOpenHelper db = NoteSQLiteOpenHelper.getInstance(getActivity());
+
```
stefan-niedermann: There's already a `NoteSQLiteOpenHelper` instance in `BaseNoteFragment`. We should better make that `protected`.

app/src/main/java/it/niedermann/owncloud/notes/util/NoteLinksUtils.java
lenzls: I agree, that it doesn't look straight forward and an explanatory comment might be needed.

**Some explanation:**

Most of the escapes are not the regex itself, but java string encoding. Java tries to interpret `\` as an escape character itself. So I need to write every `\` as `\\`.

What's left is then `\[[^\]]*\]\((\d+)\)`. Exploded this looks like
```
\[
    [^\]]*           # as much non-`]` characters as possible
\]
\(
    (\d+)            # some integer in a catch group
\)
```
I can't use `.+` in the first group, since this is evaluated greedy. So `[link](123456) and [link](654321)` would be one big match from the first `[` to the very last `]`, instead of two smaller matches.

lenzls: Still open for improvements of course :smiley:

app/src/main/java/it/niedermann/owncloud/notes/android/fragment/NotePreviewFragment.java
lenzls: agreed

app/src/main/java/it/niedermann/owncloud/notes/android/fragment/NotePreviewFragment.java
lenzls: alright, will do

app/src/main/java/it/niedermann/owncloud/notes/android/fragment/NotePreviewFragment.java
lenzls: please review again and resolve if you are happy with the changes

app/src/main/java/it/niedermann/owncloud/notes/android/fragment/NotePreviewFragment.java
lenzls: please review again and resolve if you are happy with the changes.

Ah and thanks for the review in general. Much appreciated :)

app/src/main/java/it/niedermann/owncloud/notes/util/NoteLinksUtils.java
desperateCoder: Yep, you are right. I'd suggest splitting the regex in the code like you did in your comment above and concatenate the regex, while adding these explanation comments beside the chunks. The compiler will merge the string anyway.

desperateCoder: I just looked it up, because I was wondering, why the suggested (greedy) regxex even works. In fact, it is greedy, but if the rest of the regex matches, the characters are taken away from the greedy part and the rest matches. So: you can use both expressions.

PR #647, opened 2019-12-29
#643 Empty state is shown even if there are notes

### LINKED ISSUE #643: Empty state is shown even if there are notes
I have many (184) notes on my aged smartphone (Wileyfox Swift). Loading the notes list after starting the app takes round about one second. During this time, the new empty state (#629) is shown.

This screen should be shown only if loading the list of notes from local database has been completed.

stefan-niedermann:
Fix #643 

This should display a circular progress bar while the notes list is loading and then either display the list or the empty content view.

@korelstar please review

### DIFF
--- a/app/src/main/java/it/niedermann/owncloud/notes/android/activity/NotesListViewActivity.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/android/activity/NotesListViewActivity.java
@@ -20,6 +20,7 @@ import android.view.View;
 import android.view.ViewTreeObserver;
 import android.widget.ImageView;
 import android.widget.LinearLayout;
+import android.widget.ProgressBar;
 import android.widget.RelativeLayout;
 import android.widget.TextView;
 import android.widget.Toast;
@@ -130,6 +131,8 @@ public class NotesListViewActivity extends AppCompatActivity implements ItemAdap
     RecyclerView listView;
     @BindView(R.id.empty_content_view)
     RelativeLayout emptyContentView;
+    @BindView(R.id.progress_circular)
+    ProgressBar progressBar;
 
     private ActionBarDrawerToggle drawerToggle;
     private ItemAdapter adapter = null;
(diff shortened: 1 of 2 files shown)


jancborchardt:
Very nice! :) Was about to report this as well, great to see it’s already fixed. :)

Is only a spinner shown though? It would be even nicer to show the list of notes as we know it locally directly, as was done before?

korelstar:
The list of notes is already loaded from the local database on the phone, but this takes some time, too, if there are many notes. I think both processes run in parallel: A) loading notes from the local database B) sync local database with server.

stefan-niedermann:
The spinner is shown in the time while the notes are fetched *from the local database* of the device. Depending on the amount of notes and how quick your device is, it will take physically some time to load the notes, even without any network requests.

Most people (including me) will never see the progress spinner because everything is already available offline as you said and the devices are quick enough.

jancborchardt:
> The list of notes is already loaded from the local database on the phone, but this takes some time, too, if there are many notes.

Can’t we simply prioritize the most recent, let’s say 10 notes and show them immediately, then lazyload the next ones? Because otherwise this will penalize heavy users of the Notes app.

mario:
@jancborchardt that would speed it up but it wouldn't be instant especially since the first time you do it (on app run) the DB has to open, etc, etc.

I have a similar approach to what the folks here did for Talk - spinner and then the data or empty view, depending on no. of loaded items. The reason is simple: I don't want to assume the speed of your phone.

stefan-niedermann:
Like mario said: The most time is needed for the very first time when the notes app is opened. And there a limitation to the first 10 entries will not make a difference.

**This is not a web app where we limit the amount of data because the data has to be transferred over a slow network connection.**

To be honest: I have no idea why this is a topic at all out of all sudden - The time for the first load of notes has been speeded up and is doubled as fast as it was previously according to @korelstar

I can remove the progress spinner and display a simple white page as it was before if this doesn't make you nervous while waiting :shrug:

jancborchardt:
> To be honest: I have no idea why this is a topic at all out of all sudden

The only reason is because suddenly there was an empty content view shown instead of notes on every initial load, which as @korelstar reported makes people think that data is lost when it is not. :)
I will wait for the next release and see how it is there. :) No worries @stefan-niedermann – great work!

Thanks for the info!

PR #649, opened 2019-12-29
Update Italian translation

igor-cali:


stefan-niedermann:

All translations are translated at transifex.com and if i would merge your changes here, they will be overidden at the next pull from transifex again.

Could you please provide your changes here: https://www.transifex.com/nextcloud/nextcloud/android-notes/

Issue #451, opened 2018-08-24
Toggle checkboxes in view mode


stefan-niedermann:
After #422 it is now possible to implement support for toggling checkboxes in the view mode.

There is already commented code, that does not completely work.

toxpal:
Any update on this feature? It's the only feature I miss badly about the app.

sscholl:
+1 thanks for your work!

poikilos:
@stefan-niedermann do you use the same markdown libary as the official (iOS apparently) version (take a look at this code: <https://github.com/nextcloud/notes/pull/303>)?

tobru:
Was the bug solved lately in the upstream library? If not, would you mind to link the bugreport here, so someone might be able to help with fixing.

stefan-niedermann:
Sure, this was the core problem when i tried to implement this feature: https://github.com/yydcdut/RxMarkdown/issues/55

Basically everytime when i clicked the checkbox, the markdown was presented as plain text after clicking, and not rendered.

This is also the problem for another feature (refresh in view mode). Every help is very appreciated :)

tobru:
Thanks for the pointer. Looks like the RxMarkdown projects isn't really active anymore since a long time. So IMHO this issue relates to #549 (or a follow-up issue to replace the markdown engine)

stefan-niedermann:
Yes, that's right. I am looking for alternatives, but as long as we don't have one, we are going with RxMarkdown. I will have a look at markor, as soon as i have time.

Though this does not really help for the issue. I will try to get in contact with the developer of RxMarkdown. I think (s)he might have not much time to develop, but if someone starts a PR, maybe it will get merged.

tobru:
Thanks @stefan-niedermann for taking care :+1:

quthla:
> Yes, that's right. I am looking for alternatives, but as long as we don't have one, we are going with RxMarkdown. I will have a look at markor, as soon as i have time.
> 
> Though this does not really help for the issue. I will try to get in contact with the developer of RxMarkdown. I think (s)he might have not much time to develop, but if someone starts a PR, maybe it will get merged.

https://github.com/stefan-niedermann/nextcloud-notes/issues/549#issuecomment-513489407

How about this alternative? The editor seems pretty solid by a first test

panki27:
Hey, any update on this yet? 
If I spam the checkbox in view mode (v 1.0.1) I get an 
[...]

PR #651, opened 2019-12-31
#451 Toggle checkboxes in view mode

stefan-niedermann:
[nextcloud-notes-toggle-checkboxes.apk.zip](https://github.com/stefan-niedermann/nextcloud-notes/files/4012345/nextcloud-notes-toggle-checkboxes.apk.zip)


stefan-niedermann:
> wrong checkbox is associated with action

Okay, this seems to be, because in `TextView` the `\``-charachters for code blocks are getting removed.

> refresh note using new pull-down

This is strange, never had this issue. The note gets directly saved to DB and a sync is triggered, just like if something gets changed in the `EditText`. Also with disabled network connection, this should not happen. Can you please double check? How much time passed between toggling the checkbox and pulling to refresh?

> wrong color in dark-mode

This is unrelated to this PR (happens on master, too). Maybe we can open a separate issue about this.

> different size

Yes, i noticed this, too. But not really reproducible and also on `master`. I think, it's also unrelated to this PR.

korelstar:
> > refresh note using new pull-down
> 
> This is strange, never had this issue. The note gets directly saved to DB and a sync is triggered, just like if something gets changed in the `EditText`. Also with disabled network connection, this should not happen. Can you please double check? How much time passed between toggling the checkbox and pulling to refresh?

Yes, it's reproducible. I had a look in the code and found that the local variable (belonging to the fragment) is not updated. Changing the checkbox will hence use the old text. I've fixed that in a commit to this branch.


> > wrong color in dark-mode
> 
> This is unrelated to this PR (happens on master, too). Maybe we can open a separate issue about this.

Okay, I will open a new issue for this.

stefan-niedermann:
I think i found a workaround for the "wrong checkbox is associated with action" problem. You can pull and test it :)

Issue #663, opened 2020-01-05
swipe to delete gesture is way to sensitive


paskalito:
im on a Galaxy S4 LOS Android 7

And i constantly delete notes in normal operation because the gesture is so sensitive.

I would argue that something like almost half of display with or a third of it should be actively swiped (with finger on the display) before "release animation" gets triggered / file deleted.

just for the fun:
it got much worse since https://github.com/stefan-niedermann/nextcloud-notes/issues/661 AND this https://github.com/nextcloud/server/issues/13160 affects me ^^

stefan-niedermann:
Sure, we are using `ItemTouchHelper.onSwiped()` for a consistent UX in android. So we let the framework decide how long one does need to swipe to recognize it as an actual swipe.

I did not see any configuration options (Like `Toast.LENGTH_SHORT` / `TOAST.LENGTH_LONG`), so it ended up being the default.

For me personally it always worked great and i deleted an note accidentally maybe once (which was no problem because of undo).

Maybe you have an idea how to optimize this behavior without introducing 400 lines of extra code? A PR would be very welcome :)

stefan-niedermann:
The current length depends not only on the actual dragged width but also on the speed and seems to be identical with deleting emails in Mail apps, SMS apps etc.

paskalito:
thanks for your replies!
unfortunately i don't know anything about android dev..


[TRIM_20200106_164600.mp4.zip](https://github.com/stefan-niedermann/nextcloud-notes/files/4026802/TRIM_20200106_164600.mp4.zip)

i made a video dough (with "touchpoint" enabled) to see how super sensitive it is on my device.

i don't have other apps that use this functionality besides conversations (where it is a lot less sensitive - maybe it is possible to look at that code) and opentasks (where it is actually to hard to swipe something away in my opinion)

paskalito:
https://github.com/siacs/Conversations/search?q=ItemTouchHelper.onSwiped%28%29&unscoped_q=ItemTouchHelper.onSwiped%28%29

conversations seems to import it it at least from a diffrent place

paskalito:
`getSwipeThreshold – Here we return the float value. example 0.5f means that a 50 percent swipe on the RecyclerView row would be considered as a swipe.`

from here: https://www.journaldev.com/23164/android-recyclerview-swipe-to-delete-undo

stefan-niedermann:
[...]

Issue #470, opened 2018-09-03
Ask before delete a note


kzar79:
It can happen to mistakenly delete a note (a casual touch that is interpreted as a swipe - actually it happens quite a lot to me...).

The solution is simple: before executing the deletion (line 423 of NotesListViewActivity.java), the app should ask for confirmation. 

If an additional tap is considered too troublesome for some users, this behavior could be enabled/disabled adding also a preference.

https://github.com/stefan-niedermann/nextcloud-notes/blob/7766f6504323b3e12149b0710d752c231726d83f/app/src/main/java/it/niedermann/owncloud/notes/android/activity/NotesListViewActivity.java#L423

schaarsc:
relates to #238

stefan-niedermann:
Yes, we should not ask for confirmation. There are simple ways to undo an accidential deletion: A toast to undo, a link to the trashbin from where one can easily restore a deleted note.

Issue #265, opened 2017-09-14
Easily deleted by mistake


wawont:
Easily deleted by mistake
========================
0
Brushed off and there is no note, did not have time to restore, there is no possibility to restore?

It will be better if the notes are not removed by such a simple movement.

korelstar:
duplicate of #264

Issue #238, opened 2017-08-18
Introduce a trashbin


gardiol:
Accidental delete by swipe is annoying. Yes it can be restored, but only from nextcloud file web app, and not from note app.

Please add the option to confirm swipe-delete 

This is specially needed by limited mobility or disabled people.

schaarsc:
I suggest to rename this issue into "confirm delete" because delete from menu is also missing a confirmation.

Suggestion:
- create "confirm delete" dialog with "don't show this dialog again" option
- add dialog to swipe-to-delete and menu-delete
- add confirm-delete option to preferences (this is optional, but without the don't-show-again option cannot be changed)

stefan-niedermann:
Well, i like and follow the opinion of @jancborchardt here: Undo is better than confirmation, because users are trained to simply "click away" every pop up.

So i think showing this dialog is wrong. We should better enhance swipe2delete to be less sensitive. Maybe displaying the Undo-action longer would be another point.

When we have a sidebar we could ask for extending the API to get trashbin-information so we can display a own trashbin where notes can be restored from.

schaarsc:
for me proper trashbin/undo is an acceptable alternative to confirmation dialog. toast are (depending on the situation) either shown to long or to short.

gardiol:
While in my opinion a "confirm dialog" is the only **accessible** solution, despite all the rage about material design and this (insert your adjective here) fashion... I understand that i cannot row against the masses. Nor i agree with the "let's dumb down UI because people clicks like crazy" idea... it's a make them all dumber instead of having them grow up approach which i find very damaging in the long run.

But i am not here to discuss or impose my views! so, indeed making the delay longer for the toast would help... while you implement a proper trashbin, which would be the actual solution in this case.

Or, what about a "confirmation" which is not a dialog but a button which is revealed after swiping? I am thinking at the notifications in Android 6+, where if you swipe on the persistent ones you reveal a button to access per-app notification settings. This would keep the swipe-to-delete and ensure no accidental deletions without an additional dialog? You could even remove the toast altogether!

korelstar:
I like the idea of an integrated trash bin. Is there an API for that in nextcloud/server (REST or PHP?)

stefan-niedermann:
yes i think there is a REST API, but it would list all deleted files, so the notes server app would have to provide an API that filters the entries for only deleted notes (i think this should be possible, detect if the deleted resource folder is the notes folder)

[...]

PR #665, opened 2020-01-05
Added option to not delete notes on swipe

kzar79:
This proposed change resolves an age-old problem., i.e. while scrolling notes a user accidentally deletes one with an unwanted swipe. See #663 #470 #310 #265 #225 #238
This happens way too often (at least to me), and after some damage happened to my real life because of this bug, I decided to take action.

The implemented solution is simple. A new option ("Delete note on swipe") allows the user to opt-out of the ability to delete notes by swiping. By default the option is active, meaning that the app behavior is unchanged. If the user disables the option, the swipe left is disabled (the swipe right to make a note important is not affected).

Given that there is still no integrated trashbin in the app and that it is so difficult to restore a deleted note, and that to some users (see the number of related issues) it happens very often to accidentally delete a note, this solution will make the app safer for a good amount of users.

### DIFF
--- a/app/src/main/java/it/niedermann/owncloud/notes/android/activity/NotesListViewActivity.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/android/activity/NotesListViewActivity.java
@@ -2,6 +2,7 @@ package it.niedermann.owncloud.notes.android.activity;
 
 import android.app.SearchManager;
 import android.content.Intent;
+import android.content.SharedPreferences;
 import android.content.pm.ShortcutInfo;
 import android.content.pm.ShortcutManager;
 import android.content.res.Configuration;
@@ -12,6 +13,7 @@ import android.net.Uri;
 import android.os.AsyncTask;
 import android.os.Bundle;
 import android.os.Handler;
+import android.preference.PreferenceManager;
 import android.text.TextUtils;
 import android.util.Log;
 import android.view.Menu;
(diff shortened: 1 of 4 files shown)


stefan-niedermann:
Thank you very much for your contribution. Please, before you feel offended, read the whole comment.

## I will not merge a new option, because

1. It is easy to implement it, but every new option will always cause maintenance efforts. While implementing new features, refactoring parts of the code, one will always have to check if everything still works. Even if nothing is broken, it might be function wrong how new features work in combination with existing options
2. You have mentioned a lot of issues, but in fact:

- #663 is something different (and i really would appreciate if you could write a PR **for this**.
- #470 is by you (Open the same issue multiple times doesn't count :wink: )
- #310 is about a deleted note appearing in the trashbin, we [actually voted](https://github.com/stefan-niedermann/nextcloud-notes/issues/310#issuecomment-363440669) and there was a clear weight on an `undo` option instead of a confirmation. Besides that, this was at a point in time where `swipe for favorite` has not been available yet.
- #265, #225 and #238 were at a point in time where neither `undo` nor `swipe for favorite` have been available.

## So the history reads like:

1. Many problems (which i agree)
2. `swipe for favorite` has been implemented (Reduce accidental deletion by 50%)
3. `undo` has been implemented (Easy recover)
4. Shortlink to trashbin has been implemented (and a complete trashbin support might even be implemented in the future) for everyone who misses the undo-notification and wants to restore the note at every point in time
5. **not a single issue about swipe 2 delete** for **years**
6. Bug (Mea culpa, sorry, we all are human) which caused `undo`-note to be empty
7. New issues (justified)

So no. I appreciate really really that you actually coded a PR, this is way more than most people even do. (Most come, leave a comment about how bad the world is and then go).

I just would like to improve the app for everyone and not provide 4000x options so that it fits for everyone (This will. not. work. because everyone wants something different and in the end noone understands all the trillion options).

If you want to improve the current behavior, i would suggest to contribute to #663 :rocket: 

I still wish you a happy new year and have a nice day and please, please: Do not feel offended, this is nothing personal.

PR #673, opened 2020-01-10
Fix #538 - text area covers whole screen

### LINKED ISSUE #538: Text area should cover whole screen
On devices with large screen it's not easy to tap in the small text area when creating a new note. 

I think there's no disadvantage in making it always cover whole screen (like in other note apps)

thgoebel:
Fixes #538. The EditText now expands across the entire screen. Scrolling works.

Only tested on Android 10 (OxygenOS).


stefan-niedermann:
I will test as soon as i can :) Thanks!

stefan-niedermann:
Thank you very much for your contribution! :rocket:

thgoebel:
Thank you for merging so swiftly! :)

PR #688, opened 2020-01-18
#608 Checklist auto-continuation

### LINKED ISSUE #608: Checklist auto-continuation
Hello,

I am really curious if someone already had that Idea or if someone that knows the codebase could comment on it. I think the app would gain huge on useability if check list would auto continue until you a hit enter/return key twice if the first 5 characters of the line consist of `/^[\*|-]\s\[(x|\s)\]/i`

Like:
```
* [ ] <-- return-key pressed
* [ ] (auto completed)

* [ ] <-- return-key pressed twice

(no continuation)
```

With best regards :)

stefan-niedermann:
Fixes #608

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,12 +5,11 @@ jdk: oraclejdk8
 android:
   components:
     - tools
-    - build-tools-27.0.3
-    - android-28
+    - build-tools-29.0.2
+    - android-29
     - extra-android-m2repository
 
 before_install:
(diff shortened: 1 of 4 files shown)


stefan-niedermann:
@korelstar i have some doubts regarding the performance (though i did not notice any lag). If you find some free time, could you please check if this causes performance troubles for you? I know you have one old phone :wink:

PR #694, opened 2020-01-19
#378 Enhance category handling

### LINKED ISSUE #378: Enhance category handling
When searching a category currently kind of a pop-up list appears. So it's basically a pop up in a pop up which is bad to handle.

grafik

- [x] Display the list of categories inline (expanding the pop up to the bottom), so one can easily scroll it
- [x] Display category / folder icon next to each category
- [x] Last entry should always be the search string with an `+` icon on the left instead of a folder (maybe it even should be sticky at the bottom so it is visible well)

stefan-niedermann:
Fixes #378 Enhance category handling

| Old | New |
|-----|--------|
| device-2020-01-24-212040 | alt |

For a quick review @jancborchardt since we decided in a design session that the category handling needs some enhancements :)


jancborchardt:
Very cool @stefan-niedermann! I actually meanwhile started using categories and always thought a direct list instead of the dropdown showing only when typing would be nice. :)

Issue #354, opened 2018-04-01
Password protection


awolfman:
Is password protection supposed?

kasinjsh:
I agree, more security is needed. For example I hold important information in notes amd app have no protection.

stefan-niedermann:
> Best would be unlocking using device credentials

Yes, i agree that we should adapt this from the Nextcloud Files app (without the option to use a own key, but using the device login).

This issue has currently not a high priority for me, but you can start a PR or put a bounty on it if you are not a developer :rocket:

stefan-niedermann:
Update: I did a proof of concept which basically works but has a problem with `PendingIntent`s which basically means, that the app is not able to open a note from a widget
If any android dev is here around wanting this feature, it would be awesome if he could help me out a little bit in this branch: https://github.com/stefan-niedermann/nextcloud-notes/pull/704

stefan-niedermann:
Fixed in version `2.11.0`.

PR #704, opened 2020-01-25
#354 Password protection

stefan-niedermann:
Proof of concept

Noticed problem: With enabled password protection, SingleNoteWidgets can no longer be opened. ~~Opening a note from the NotesListWidget works flawlessly.~~ Can you maybe have a look at this special case @dan0xii? The widgets are still a mystery for me with all the PendingIntent magic 😅😅


dan0xii:
Sure. They are not much clearer to me !

stefan-niedermann:
Did you have any time to have a look at this yet, @dan0xii ? 🙂

stefan-niedermann:
@dan0xii A bounty of 50$ has been added to this issue. You can have it if you are able to fix the widgets :laughing:

beatbrot:
May I ask what exactly is not working in this branch?

I can click SingleNoteWidgets without problem on my end.

Just to get things right, my workflow is:

1. Enable password protection
2. Create single note widget
3. Optional: Kill App
4. Click on single note widget.

stefan-niedermann:
Step 5: After submitting the correct PIN / finger print, the app will simply completely close instead of displaying the app.

dan0xii:
Re: https://github.com/stefan-niedermann/nextcloud-notes/pull/704/commits/804c8ac58411bfa685bc0870cc3e67759e3fe890. FLAG_ACTIVITY_NO_HISTORY was there originally to avoid confusing the user when multiple notes were opened from the widget. You'd press the back button and be presented with notes that you weren't expecting.

beatbrot:
@dan0xii Unfortunately I do not completely understand what you mean... Could you provide me steps to reproduce the problem you mentioned?

beatbrot:
I am also wondering in which case [these lines](https://github.com/stefan-niedermann/nextcloud-notes/pull/704/files#diff-3784dc03c6dd3823f74b322bc847fd99R39-R41) actually matter.

```java
    @Override
    protected void onResume() {
        super.onResume();

        if (!isTaskRoot()) {
            askToUnlock();
        }
    }
```

Maybe someone can explain it to me :)

stefan-niedermann:
@beatbrot i think they are affected when you open the app and go to a note (and switch preview/edit fragments)...

oooor maybe when creating a note from the notification bar or the create widget, don:t remember exactly...

stefan-niedermann:
@dan0xiiaybe you have a better idea baout how this could be handled?

beatbrot:
> 
> 
> @beatbrot i think they are affected when you open the app and go to a note (and switch preview/edit fragments)...
> 
> oooor maybe when creating a note from the notification bar or the create widget, don:t remember exactly...

Ah I think I could fix many of those cases by moving the `askForUnlock()`-call from `onCreate()` to `onStart()`. (`onStart()` being the method that is called as soon as the activity gets visible)

dan0xii:
The code as of https://github.com/stefan-niedermann/nextcloud-notes/commit/ca718370 appears to work for me. The single note widget asking you to enter your device pin and then displaying the note when you do so.


> > @beatbrot i think they are affected when you open the app and go to a note (and switch preview/edit fragments)...
> > oooor maybe when creating a note from the notification bar or the create widget, don:t remember exactly...
> 
> Ah I think I could fix many of those cases by moving the `askForUnlock()`-call from `onCreate()` to `onStart()`. (`onStart()` being the method that is called as soon as the activity gets visible)
> 
> See commit [dd36786](https://github.com/stefan-niedermann/nextcloud-notes/commit/dd36786a8184cc06d72db43ebc95b1ac3c4d9a66)

@stefan-niedermann is right, we still need the call to askToUnlock() in onResume to cover scenarios where the user navigates between this app and another.

I don't think there is a need to move askForUnlock to onStart() because it's either called the first time the activity is created in onCreate() or when the user navigates between the app and another app, or their homescreen, in onResume().

beatbrot:
@dan0xii I am not sure whether we really can leave it in `onCreate()`.

As you can see in the video, I am not asked for a confirmation when I open the second note (this is as of commit 8d3e9c22).



In the most recent version (with the check in `onStart()`) this doesn't happen and I get asked for fingerprint twice.

dan0xii:
I've replicated the problem from your video in https://github.com/stefan-niedermann/nextcloud-notes/commit/8d3e9c223bcd2393de19df5a1356e0b587a2f39e and I see that onResume in LockedActivity calls askToUnlock() which then exits because Notes.isLocked() is false. See [here](https://github.com/stefan-niedermann/nextcloud-notes/pull/704/files#diff-3784dc03c6dd3823f74b322bc847fd99R74):

```java
    private void askToUnlock() {
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP && Notes.isLocked()) {
            KeyguardManager keyguardManager = (KeyguardManager) getSystemService(Context.KEYGUARD_SERVICE);
            if (keyguardManager != null) {
                Intent i = keyguardManager.createConfirmDeviceCredentialIntent(getString(R.string.unlock_notes), null);
                i.addFlags(Intent.FLAG_ACTIVITY_SINGLE_TOP);
                startActivityForResult(i, REQUEST_CODE_UNLOCK);
            } else {
                Log.e(TAG, "Keyguard manager is null");
            }
        }
```

The app is not locked because onStop in LockedActivity didn't get past isTaskRoot(). Hope that makes sense. I'm just testing out a fix.

dan0xii:
@stefan-niedermann, can you clarify how password protection should work. 

The prompt to unlock should be displayed when:
1. The user opens the app
2. The user returns to the app after switching to another
3. Any note is opened via widget

There should be no difference as to whether the user has a note open in the app or not in the above scenarios, the unlock prompt is displayed ?

stefan-niedermann:
Basically the propt should be displayed, when the user opens any UI (app **or** widget).

But when the UI is already open and the user navigates, e.g. from the list to edit or view a single note), no prompt should be shown of course.

The app should be locked again, if the app hasn't been used for some time (we don't need to set an exact timer here but rely on `onStop`).

beatbrot:
@stefan-niedermann 

What do you think about using a fixed amount of time anyways? Like say 5/10/15 seconds?

I totally see users wanting this when they are switching back and forth between e.g. a web browser and their notes app.

stefan-niedermann:
> fixed amount of time

I really wanted to avoid this if possible. Tested the current behavior for some time and except the widgets i was happy :smile:

beatbrot:
Alright, I understand :)

dan0xii:
> We don't need to set an exact timer here but rely on `onStop` because this allows users to still enter the UI if they switched to another app for a short time.

Understood. So if the user has been into the app, unlocked it and then opens a note from the homescreen they shouldn't be prompted to unlock. It would need onStop to occur to lock it again?

I'm thinking of a situation where the user opens multiple notes from the single note widgets or from the note list widget--we're allowing them the 'grace period' before the app goes to sleep and onStop is called.

stefan-niedermann:
Exactly 👍👍👍

beatbrot:
With that knowledge, putting the check in onStart now really just makes unnecessary trouble. Will revert that commit

dan0xii:
I'm testing on android 7.1.1, just focusing on the locking behaviour, and if I open the app from 'cold', after unlocking it, I get the following:

I open a note and see onStop is called by LockedActivity and isTaskRoot returns True therefore Notes.lock is called. I go back to the main note list with no unlock prompt but then I must unlock to open another note (or the Settings menu).

Has this behaviour been introduced unknowingly or is there a problem with my testing?

stefan-niedermann:
@dan0xii can you test with thsi commit: https://github.com/stefan-niedermann/nextcloud-notes/commit/fee05a32cc8002bd9f61bbfa130db5f55407d95a

If i remember correctly, this behavior wasn't there at this time

beatbrot:
Just tested it with fee05a3. The problem still occurs.

By the way: An implementation with a fixed timeout would fix this problem. Every back press (Note -> NoteList) and every start of an activity (NoteList -> Note) would reset this timer and therefore allow the user to always navigate the app (well at least this is true for Note and NoteList activites).

dan0xii:
There's that or we could move to onDestroy() which I think would still cover off the intended scenarios.

beatbrot:
> 
> 
> There's that or we could move to onDestroy() which I think would still cover off the intended scenarios.

If this fixes the problem, then I think it would be the better solution :)

stefan-niedermann:
What's the lifecycle of the timer though? If we do not store it in SharedPrefs, it will be resetted on each app kill, right (like it is now)?

And you want to reset the timer each time an activuty interaction appears?

But still: If we navigate within the app, there should be no prompt which interrupts the user. Even if one is on the list view for 10 minutes (exceeds timer) and *then* opens a note.

beatbrot:
The timer would be a static variable in the Notes-Class so it would live as long as the application does.

Opening activities after 5h shouldn't be a problem because we would just override the `startActivity`-method so that it resets the timer just before staring the actual activity.

stefan-niedermann:
Okay, let's try it with a timer then.

stefan-niedermann:
I would propose to start with 30 seconds for the beginning (we can adjust the value later if it's insufficient).

dan0xii:
Did a rebase just go slightly awry? :)

I'll continue testing with onDestroy() in [dan0xii/354-password-protection.dan0xii](https://github.com/dan0xii/nextcloud-notes/tree/354-password-protection.dan0xii). It still needs the Settings activity tweaking but the widgets and main app seem to test fine.

beatbrot:
> 
> 
> Did a rebase just go slightly awry? :)
> 

I don't really know what you mean :) Could you explain?

beatbrot:
Btw, I just commited the timer based implementation :)

stefan-niedermann:
Good enough for a beta - let's get that change in and then let's hear what the users on the beta-channel and on F-Droid say. Thank you, both.

Sorry @dan0xii but this bounty goes to @beatbrot :man_shrugging:

PR #705, opened 2020-01-26
Back button on notes opened from other notes

lenzls:
Intended to fix back button behaviour on notes that were opened from other notes (see #646) as reported in [this comment](https://github.com/stefan-niedermann/nextcloud-notes/issues/623#issuecomment-578415804).

To be honest, I'm not 100% sure what is going on. Just removing the activity close code on `onBackButtonPressed` wasn't enough. Creating new fragments instead of activities seems to fix the issue. Still I'd be happy to have some discussion on the code.

I did some manual testing and didn't notice weird side effects by the changes.


stefan-niedermann:
Closing this PR for now because it is quite outdated (conflicts with `master`).

Sorry, it is totally my fault that i didn't look into this earlier, but i will check the changes again as there are only a few lines affected.

PR #713, opened 2020-02-01
Fix #679 Lots of empty notes are created if the server is full

### LINKED ISSUE #679: Lots of empty notes are created if the server is full
**Describe the bug**
I recently tried to create a new note while my Nextcloud server was full. Every time the app tried to sync, dozens of empty notes were added. The problem disappeared once I cleared some space from my Nextcloud server. 

I don't know if it is relevant but I don't have my own server, I have an account with limited storage from a 3rd party provider. 

**To Reproduce**
1. Log in to an account with no space left
2. Create a new note and sync it. 

**Server**
 - Nextcloud-Version: 17.0.0.9
 - Nextcloud Notes-Version (server app): unknown 
 - PHP-Version: unknown

**Smartphone (please complete the following information):**
- Device: Sony Xperia x compact 
 - Android-Version: 8.0.0

stefan-niedermann:
Fixes #679 

Signed-off-by: stefan-niedermann <info@niedermann.it>

### DIFF
--- a/app/src/main/java/it/niedermann/owncloud/notes/persistence/NoteServerSyncHelper.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/persistence/NoteServerSyncHelper.java
@@ -364,6 +364,8 @@ public class NoteServerSyncHelper {
                 } catch (NextcloudHttpRequestFailedException e) {
                     if (e.getStatusCode() == 304) {
                         Log.d(TAG, "Server returned HTTP Status Code 304 - Not Modified");
+                    } else if(e.getStatusCode() == 507) {
+                        Log.d(TAG, "Server returned HTTP Status Code 507 - Insufficient Storage");
                     } else {
                         exceptions.add(e);
                     }
@@ -423,6 +425,9 @@ public class NoteServerSyncHelper {
                 Log.d(TAG, "Server returned HTTP Status Code " + e.getStatusCode() + " - " + e.getMessage());
                 if (e.getStatusCode() == 304) {
                     return LoginStatus.OK;
+                } else if(e.getStatusCode() == 507) {
+                    exceptions.add(e);
+                    return LoginStatus.INSUFFICIENT_STORAGE;
                 } else {
                     exceptions.add(e);
                     return LoginStatus.JSON_FAILED;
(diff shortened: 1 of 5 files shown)


app/src/main/java/it/niedermann/owncloud/notes/persistence/NoteServerSyncHelper.java
```diff
@@ -364,13 +367,21 @@ private void pushLocalChanges() {
                 } catch (NextcloudHttpRequestFailedException e) {
                     if (e.getStatusCode() == 304) {
                         Log.d(TAG, "Server returned HTTP Status Code 304 - Not Modified");
+                        return LoginStatus.OK;
```
korelstar: What happens if there where multiple changes? `db.getLocalModifiedNotes` will return multiple notes and the for-loop is executed multiple times. However, in some cases, the execution is cancelled due to these new return statements. This means, all local changes that are behind a local change with error will not be executed.

An interesting use case is if the storage is full. I change a note and see the error message. Then I want to delete old notes in order to free some storage. However, the deletion will never be executed, since `pushLocalChanges` is cancelled after the first error.

stefan-niedermann:
You are totally right. What do you propose? Make a `LoginStatus`-variable outside of the `for`-loop with default value `OK` and assign another value in error case?

Of course each iteration could theoretically produce still different errors and this would only allow one to see the very last message. But then a user could work on one error after another.

Thank you for your review btw :)

korelstar:
That approach would be better. Although it wouldn't show all different errors, I think that would be fine.

Another option would be to build a `Set<LoginStatus>` and to add every status which is not `LoginStatus.OK`. Later, a string could be builded by mapping the elements to their error message and concatenate them together. But that could be confuse the user.

stefan-niedermann:
:laughing: i really would be so mad if i would like to sync a bunch of notes and then immediately would receive a toast (!) which covers the whole screen and contains 7 error messages :laughing: 

No, indeed i think it's actually a feature to display only one error message at the time.

Will rewrite this part and adjust to what we discussed.

stefan-niedermann:
Have a look at #721

PR #721, opened 2020-02-04
Do not break push on error

stefan-niedermann:
Follow-up of #713


stefan-niedermann:
Great :) Did you have a look at the comment i added regarding the "recreate of note which has been deleted on the server"?

I think this failed silently since some time  already.

The server issue about the `404` status code should fix this, i will add the code without testing it and will have a look at it as soon as you released the next version of the server with the fix.

korelstar:
Yes, you just have to add a try-catch around `notesClient.editNote` and check if the Exception origin is HTTP 404. Then re-throw the exception if it is not.

korelstar:
Oh, and you should do this for deleted in the same way!

stefan-niedermann:
@korelstar have a look at [my latest commit](https://github.com/stefan-niedermann/nextcloud-notes/pull/721/commits/e3f69482c7e8df2619f0fa978adbc5b1cc285df4). I adjusted the code slightly, but the documentation (in the `Log.v()`s) should be self-explaining.

Did i miss anything obvious?

PR #742, opened 2020-02-24
Remove requirement for physical device/emulator to run tests

beatbrot:
# Description
By moving the tests from `androidTest` to `test`, the test cases can be executed on the host machine and therefore do not need an emulator or a physical device to run them.

### Remarks

I had to re-implement String#join because we can neither use `android.text.TextUtils` nor `String#join`.


stefan-niedermann:
Thank you for your contribution! 😊

I don't know much about testing with android, and your PR sounds reasonable. But i can't imagine that `TextUtils` cannot be used without an actual physical device, huh?

I am not friend of adjusting productive code only for testing reasons, especially in this case we would draw a whole method in our project just to satisfy tests (and we should test the method itself, too 🤔).

Why can't we use `String.join` though?

beatbrot:
Thanks for your very quick reponse! I'll try to address your remarks as good as possible:

#### Why can't we use `TextUtils`?
As soon as we put the tests inside the `test`-folder, they will be ran on the JDK that is installed on the local pc. Therefore, we do not have access to the Android API (since we have no actual device). By default, a mock of the android.jar (containing the Android API) is used, that throws an exception on every method call. This can be changed so that all methods just return `null` when called. Both settings wouldn't work for us since we rely on the returned object. 

What is additionally troubling is, that `TextUtils#join` is static and therefore not easily mockable.

#### Why can't we use `String.join(...)`?

Unforunately, `String#join(...)` (as well as `StringJoiner`) were introduced in JDK 1.8 which is only available since Android API Level 24. The note apps still targets 14, so increasing minAPI to 24 seems rather drastic.


### Conclusion

Overall, I agree, that modifying production code to enable testability seems wrong. Personally, I'd let it slide on this occasion since the join-method is not really that complex and also doesn't appear as a public member. But I understand, that the final decision is up to you. 

We could also leave `NoteLinkUtilsTests` in `androidTest` and just move the rest of the tests to `test`.

I am happy to hear your opinion on this

stefan-niedermann:
Thank you for your information :)

> We could also leave NoteLinkUtilsTests in androidTest and just move the rest of the tests to test.

That sounds nice, let's do this as a first step and i will have a look what's possible.

Does the GitHub workflow needs to be adjusted or is the execution command the same? https://github.com/stefan-niedermann/nextcloud-notes/blob/master/.github/workflows/android.yml#L16

About JDK 1.8: I thought this was configured in the [build.gradle](https://github.com/stefan-niedermann/nextcloud-notes/blob/master/app/build.gradle#L8) and in the [workflow](https://github.com/stefan-niedermann/nextcloud-notes/blob/master/.github/workflows/android.yml#L14) and the build process does some magic desugaring so one can use at least a subset of JDK 1.8 on lower Android target versions?

beatbrot:
I made the changes you suggested.

> Does the GitHub workflow needs to be adjusted or is the execution command the same? https://github.com/stefan-niedermann/nextcloud-notes/blob/master/.github/workflows/android.yml#L16

Well this one is kinda weird. Since `androidTest`-tests need an android device, it is hard to run them inside Github Actions. Additionally, `gradle test` afaik does only run `test`-tests but not `androidTest`-tests. So by using the new model, the command will actually have *some* effect compared to before. 

> About JDK 1.8: I thought this was configured in the build.gradle and in the workflow and the build process does some magic desugaring so one can use at least a subset of JDK 1.8 on lower Android target versions?

Unfortunately I am not too experienced in this area. But you can just try using `String.join()` in the source code. Maybe you understand the error message better than me :)

stefan-niedermann:
Thank you :) I had time to check this topic out a little bit (now i am at a pc :wink: ) and understand now what the problem with the `TextUtils` / `StringUtils` is. :)

Looking forward to further contributions :rocket:

stefan-niedermann:
I found a solution. `JUnit` can "overwrite" classes for tests, if they are located correctly. You can have a look at the commit https://github.com/stefan-niedermann/nextcloud-notes/commit/75bdb4c2cca69609e520e8c97f7d53489287dece if you are interested.

beatbrot:
Interesting, didn't know that! Good job :)

PR #744, opened 2020-02-24
Add "System default" to darkmode settings

beatbrot:
# Description

As discussed way back in #304, currently  the app is missing an option to follow the global setting for darkmode. This PR aims to fix this by introducing said option.

The change is rather large and hard to unit test so it would be nice if someone could look over it/test it on an emulator/device.

### DIFF
--- a/app/src/main/java/it/niedermann/owncloud/notes/android/DarkModeSetting.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/android/DarkModeSetting.java
@@ -16,6 +16,8 @@ import java.util.NoSuchElementException;
  * @see AppCompatDelegate#MODE_NIGHT_FOLLOW_SYSTEM
  */
 public enum DarkModeSetting {
+    // WARNING - The names of the constants must *NOT* be changed since they are used as keys in SharedPreferences
+
     /**
      * Always use light mode.
      */
--- a/app/src/main/java/it/niedermann/owncloud/notes/android/DarkModeSetting.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/android/DarkModeSetting.java
@@ -1,0 +1,64 @@
+package it.niedermann.owncloud.notes.android;
+
+import androidx.appcompat.app.AppCompatDelegate;
+
+import java.util.NoSuchElementException;
+
+/**
+ * Possible values of the Dark Mode Setting.
+ * <p>
+ * The Dark Mode Setting can be stored in {@link android.content.SharedPreferences} as String by using {@link DarkModeSetting#name()} and received via {@link DarkModeSetting#valueOf(String)}.
+ * <p>
+ * Additionally, the equivalent {@link AppCompatDelegate}-Mode can be received via {@link #getModeId()}. To convert a {@link AppCompatDelegate}-Mode to a {@link DarkModeSetting}, use {@link #fromModeID(int)}
+ *
+ * @see AppCompatDelegate#MODE_NIGHT_YES
+ * @see AppCompatDelegate#MODE_NIGHT_NO
+ * @see AppCompatDelegate#MODE_NIGHT_FOLLOW_SYSTEM
+ */
+public enum DarkModeSetting {
+    /**
+     * Always use light mode.
+     */
+    LIGHT(AppCompatDelegate.MODE_NIGHT_NO),
+    /**
+     * Always use dark mode.
+     */
+    DARK(AppCompatDelegate.MODE_NIGHT_YES),
+    /**
+     * Follow the global system setting for dark mode.
+     */
+    SYSTEM_DEFAULT(AppCompatDelegate.MODE_NIGHT_FOLLOW_SYSTEM);
+
+    private final int modeId;
+
+    DarkModeSetting(int modeId) {
+        this.modeId = modeId;
+    }
+
+    public int getModeId() {
+        return modeId;
+    }
+
+    /**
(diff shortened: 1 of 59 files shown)


stefan-niedermann:
Nice!! That's indeed a big change. Maybe @korelstar or @dan0xii can review and test it, too?

I will need some time to review the changes.

korelstar:
I'll try to do a review during this week ...

stefan-niedermann:
Fresh install of the latest build:

1. Open app the very first time
2. Import account
3. Go to settings
4. See "System default" set
5. Change to "Light" - **crash**

```
App Version: 2.10.3
App Version Code: 81
Files App Version Code: 30100190

---

OS Version: 4.9.185-perf-g301928a(5d3e3c161a)
OS API Level: 28
Device: beryllium
Model (and Product): POCO F1 (beryllium)

---

java.lang.IllegalArgumentException: No enum constant it.niedermann.owncloud.notes.android.DarkModeSetting.Hell
	at java.lang.Enum.valueOf(Enum.java:258)
	at it.niedermann.owncloud.notes.android.DarkModeSetting.valueOf(DarkModeSetting.java:18)
	at it.niedermann.owncloud.notes.android.fragment.PreferencesFragment.lambda$onCreatePreferences$0$PreferencesFragment(PreferencesFragment.java:33)
	at it.niedermann.owncloud.notes.android.fragment.-$$Lambda$PreferencesFragment$8qg98-aBeAUBtVFGdlXTbTa_lhM.onPreferenceChange(Unknown Source:2)
	at androidx.preference.Preference.callChangeListener(Preference.java:1118)
	at androidx.preference.ListPreferenceDialogFragmentCompat.onDialogClosed(ListPreferenceDialogFragmentCompat.java:107)
	at androidx.preference.PreferenceDialogFragmentCompat.onDismiss(PreferenceDialogFragmentCompat.java:267)
	at androidx.fragment.app.DialogFragment$3.onDismiss(DialogFragment.java:116)
	at android.app.Dialog$ListenersHandler.handleMessage(Dialog.java:1393)
	at android.os.Handler.dispatchMessage(Handler.java:106)
	at android.os.Looper.loop(Looper.java:193)
	at android.app.ActivityThread.main(ActivityThread.java:6718)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)

```

stefan-niedermann:
There seems to be an issue with the preference value - the value is translated (this is actually also in master the case as far as i know) - we should tackle this issue and migrate all currently translated theme *values* to a hard coded english string.

beatbrot:
Seems like the ENUM Constant "LIGHT" was wrongfully translated to "Hell". This shouldn't happen. I should look into that.

Also, since the old setting was saved as an `boolean` in `SharedPreferences` and now has to be saved as String, we also need a migration method for that (since we use `SharedPreferences#getString(DARK_MODE, ...)` all over the place). I am not sure yet what a good solution for that might be...

stefan-niedermann:
Oh, i messed this up with another setting.

Though yes, the value is translated and shouldn't.

The migration path from the previous boolean value seems to be pretty obvious to me:

- `true` -> dark theme
- `false` -> light theme
- unset -> system default

What do you think about using an fake database upgrade to migrate the values (an delete the previous boolean values from the shared preferences after the migration)?

beatbrot:
I have to be honest, I never heard about the concept of a database upgrade :) I'd have to look into that.

stefan-niedermann:
The idea is that a user with an very old .apk can easily upgrade without loosing all his settings and stuff, even if the tables changed.

You can have a look here: https://github.com/stefan-niedermann/nextcloud-notes/blob/master/app/src/main/java/it/niedermann/owncloud/notes/persistence/NoteSQLiteOpenHelper.java#L152

At some places in the `onUpgrade`-method we are already modifying shared preferences.

beatbrot:
Ah, now I understand! We do not really do a database upgrade but instead fake one to get the chance to fix up SharedPreferences?

beatbrot:
I added some migration logic for old users. Unforunately, the database migration happens after the first time the SharedPreferences are read. Therefore I still had to handle these cases :/

beatbrot:
@korelstar I tried to incorporate your suggetions :) Maybe you can take another look at it?

korelstar:
Thanks! A tiny optimization would be to move the definition of `DARK_THEME_KEY` to `NoteWidgetHelper` and to remove the parameter `darkModeKey` from `NoteWidgetHelper.getDarkThemeSetting`, since it's always the same (replace its usage with `DARK_THEME_KEY`).

beatbrot:
@korelstar It actually is NOT always the same. In [SingleNoteWidget](https://github.com/stefan-niedermann/nextcloud-notes/blob/482be9b0931bc002086141a2dd4ae1c3bb6e069f/app/src/main/java/it/niedermann/owncloud/notes/android/appwidget/SingleNoteWidget.java#L37), it resolves to `SNW_darkTheme`, while it resolves to "NLW_darkTheme" in [NoteListWidgetFactory](https://github.com/stefan-niedermann/nextcloud-notes/blob/482be9b0931bc002086141a2dd4ae1c3bb6e069f/app/src/main/java/it/niedermann/owncloud/notes/android/appwidget/NoteListWidgetFactory.java#L39)

korelstar:
Ah, okay. Then it's fine, of course! :-)

PR #789, opened 2020-04-15
Correction of the description

Valdnet:
Word correction "hesitate".


stefan-niedermann:
Thank you, but all our translations are done on [transifex](https://www.transifex.com/signin/?next=/nextcloud/nextcloud/android-notes/) by the community.

If we change it here, they will be overridden the next time we sync the translations, so please update the translation at transifex :slightly_smiling_face:

Valdnet:
> Thank you, but all our translations are done on transifex by the community.

@stefan-niedermann  Yes I know about it. I translate at Transefix (I'm a reviewer) at Nextcloud 😉 .

> If we change it here, they will be overridden the next time we sync the translations, so please update the translation at transifex 🙂

In my opinion, this should be changed here so that the English version of the description is displayed correctly. Translators will translate them again in their language.

stefan-niedermann:
Oops, you are absolutely right! Sorry, i thought this was about a *translated* language, didn't recognize that it is english.

PR #802, opened 2020-04-21
Bug fix: 769

Isaac-Graham:
The main problem is that it costs too much time in the function of countOccurences. The time complexity of java library function indexOf is m*n (m is the length of the pattern and the n is the length of the search string). So that the worst time complexity of countOccurences is O(mn^2), which is too large. Because the main logic is running in the main thread so that the UI will be frozen for a long time. We use a O(m+n) implementation, which will solve this issue.

We have tested both in Huawei P30 pro and AVD. In our test, we create three test notes, the length of which are 28672, 114688 and 229376 separately.
In Huawei P30 pro, with using the indexOf, the average time are 140ms, 520ms, 820ms separately.
In AVD, with using the indexOf, the average time are 1039ms, ANR, ANR.
However, with using our implementation, the average times are significantly reduced.
In Huawei P30 pro, with our implementation, the average time are 1ms, 1ms, 2ms.
In AVD, with our implementation, the average time are 2ms, 9ms, 24ms.

### DIFF
--- a/app/src/main/java/it/niedermann/owncloud/notes/android/fragment/SearchableBaseNoteFragment.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/android/fragment/SearchableBaseNoteFragment.java
@@ -216,6 +216,9 @@ public abstract class SearchableBaseNoteFragment extends BaseNoteFragment {
     }
 
     private static int countOccurrences(String haystack, String needle) {
+        if (haystack == null || haystack.isEmpty() || needle == null || needle.isEmpty()) {
+            return 0;
+        }
         Matcher m = Pattern.compile(needle, Pattern.CASE_INSENSITIVE | Pattern.LITERAL)
                 .matcher(haystack);
 
--- a/app/src/main/java/it/niedermann/owncloud/notes/android/fragment/SearchableBaseNoteFragment.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/android/fragment/SearchableBaseNoteFragment.java
@@ -1,6 +1,7 @@
 package it.niedermann.owncloud.notes.android.fragment;
 
 import android.os.Bundle;
+import android.os.Handler;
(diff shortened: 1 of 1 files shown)


stefan-niedermann:
For similar szenarios (e. g. validation of input fields, save debouncing when editing, etc...) often a debounce time of round about 200ms is used. Maybe you want to try this?

app/src/main/java/it/niedermann/owncloud/notes/android/fragment/SearchableBaseNoteFragment.java
```diff
@@ -104,6 +108,10 @@ public void onGlobalLayout() {
         }
 
         searchView.setOnQueryTextListener(new SearchView.OnQueryTextListener() {
+            private final int delay = 20; // If the search string does not change after $delay ms, then the search task starts.
```
stefan-niedermann, line 111: this should be a `private static final` constant at the top of this class

app/src/main/java/it/niedermann/owncloud/notes/android/fragment/SearchableBaseNoteFragment.java
```diff
@@ -123,7 +136,36 @@ public boolean onQueryTextChange(String newText) {
...
-                return true;
+            }
+
+            private void queryWithHandler(String newText) {
+                if (delayQueryTask != null) {
+                    delayQueryTask.cancel();
+                    handler.removeCallbacksAndMessages(null);
+                }
+                delayQueryTask = new DelayQueryRunnable(newText);
+                handler.postDelayed(delayQueryTask, delay);
```
stefan-niedermann, line 147: This should only be delayed if the `newText` length is > 1. Otherwise typing in the first character of the search is also delayed and this feels strange. It should immediately search for the first entered character and only delay all further searches.

PR #805, opened 2020-04-24
804: update complete search result when submit or prev is clicked.

mikechesterwang:
1. 804: Call `colorWithText` to highlight the result when `submit` is clicked.
2. While searching some word, if the same word is appended or removed from the content, `occurrenceCount` is not updated and `prev` button cannot act normal. So update `occurrenceCount` when `prev` is clicked.


stefan-niedermann:
Good job :+1:

PR #810, opened 2020-04-30
Files -> Android

kesselb:
As per https://github.com/stefan-niedermann/nextcloud-deck/pull/442


stefan-niedermann:
Also here there are some `.txt`-files in zhe `fastlane` folder which contain this wording and should be adjusted for uniformity.

PR #815, opened 2020-05-06
Normalize database

Isaac-Graham:
Previously, the database has only two tables `NOTES` and `ACCOUNT`, and the category information is written in each note. In order to normalize the database, this pr extract the `CATEGORY` from `NOTES`. This pr has been tested a lot by ourselves and I hope that it is bug-free. However, as you can see, this pr modifies a large number of codes in the database. So that it might be some cases that we have not considered. If there is any suggestion, no matter finding a bug or something else, please do not hesitate to contact me. I will definitely try to fix them.

In the `AbstractNotesDatabase.java`, this pr create a new table called `CATEGORIES` with three columns, which are `ID`, `ACCOUNT_ID`, and `TITLE`. The `ID` is used as a foreign key in the table `NOTES` and the `ACCOUNT_ID` is a foreign key from the table `ACCOUNTS`. We also update the database version to `13` and we move data from the old database to the new database, which is written in `onUpgrade()`.

In `NotesDatabase.java`, there are a large number of modifications. The main idea for these modifications is not to change the interface for the public methods written. We modified some methods originally using `category` in `NOTES`. The only thing needed to be done is to change the type of `category`, which is originally String, to Integer. We also modified the methods to get categories. Because the structure of category storage is completely changed, theses methods about categories are overridden by us.

At the same time, in the previous [pr](https://github.com/stefan-niedermann/nextcloud-notes/pull/802), I find that the delay is more a static attribute than an instance attribute. This modification is also done in this pr. And I also write some comment on the previous pr.

The test files both for this pr and the last pr are also committed. These test files are written by my teammates. Thanks to @WingsUpete and @Hui-Ouyang16.

Hope that this pr is bug-free. We add some comment and JavaDoc for all most all of the modifications in this pr. If there is any question, please contact me immediately.


app/src/main/java/it/niedermann/owncloud/notes/persistence/NotesDatabase.java
```diff
@@ -225,15 +227,15 @@ private DBNote getNoteFromCursor(long accountId, @NonNull Cursor cursor, boolean
...
         return new DBNote(
                 cursor.getLong(0),
                 cursor.getLong(1),
                 modified,
                 cursor.getString(3),
                 pruneContent ? "" : cursor.getString(9),
                 cursor.getInt(5) > 0,
-                cursor.getString(6),
+//                cursor.getString(6),
+                getTitleByCategoryId(accountId, cursor.getInt(6)),
```
korelstar, line 238: This will slow down the down the app, since you are executing an additional SQL query for every note. Currently, I have about 200 notes and showing them on the main Activity requires a single SQL query. With this change, 201 SQL queries are executed. This is a "no go"!

There are two alternatives to fix this issue:
1. build a "cache" in memory for categories, i.e. load all categories of the current account to a `HashMap` on creation of e.g. `NotesDatabase` and query this cache in `getTitleByCategoryId(...)` (and related methods). But on every change of categories (insertion, deletion, edit), you must update this cache.
2. change the SQL query in `getNotesCustom(...)` and add a JOIN on the categories table. Hint: you can reduce the need for changing code if you change the column names of the categories table: If they differ from column names in the notes table, you may not have to prepend the table name (e.g. `CATEGORIES.TITLE`, because `TITLE` exists in both tables `CATEGORIES` and `NOTES`).

I vote for the second approach. You can remove the method `getTitleByCategoryId(...)`, then.

app/src/main/java/it/niedermann/owncloud/notes/persistence/NotesDatabase.java
```diff
@@ -769,4 +842,102 @@ private static void validateAccountId(long accountId) {
...
+                key_id,
+                null,
+                key_id);
+        int id = -1;
+        if (cursor.moveToNext()) {
+            id = cursor.getInt(0);
+        }
+        cursor.close();
+        return id;
+    }
```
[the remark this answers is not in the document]
Isaac-Graham: Because I found that the return value of `addCategory()` is the row ID of the newly inserted row. But the row ID I think is sth different from the primary key ID? I don't know if I got it wrong but if you can make sure that the return value of `addCategory()` is exactly the primary key ID,  I will do it.

Isaac-Graham:
For removing obsolete categories, what about doing it in `setCategory(...)`? I think the only one possibility to trigger "obsolete categories" is that the note's category is set to another one. So that we can query in db to find out whether there is such a category without any note and remove it.

app/src/main/java/it/niedermann/owncloud/notes/persistence/NotesDatabase.java
```diff
@@ -303,10 +309,29 @@ private DBNote getNoteFromCursor(long accountId, @NonNull Cursor cursor, boolean
...
+            for (int i = 0; i < ids.size(); i++) {
+                tmpSQL.append(tmp);
+                if (i != ids.size() - 1) {
+                    tmpSQL.append(" or ");
+                }
+            }
+            where.add(String.format("( %s )", tmpSQL.toString()));
+            for (int i = 0; i < ids.size(); i++) {
+                args.add(String.valueOf(ids.get(i)));
+            }
```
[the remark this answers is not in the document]
Isaac-Graham: oooooops, I forgot to use subquery....I will do it later.

app/src/main/java/it/niedermann/owncloud/notes/persistence/NotesDatabase.java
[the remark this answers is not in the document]
korelstar: According to the answers on Stackoverflow, it is the same. But we will see if it's working or not ...

korelstar:
> For removing obsolete categories, what about doing it in `setCategory(...)`? I think the only one possibility to trigger "obsolete categories" is that the note's category is set to another one. So that we can query in db to find out whether there is such a category without any note and remove it.

That's not sufficient. There are updates of the category column at least also in `updateNoteAndSync` and `updateNote` (can be called during synchronisazion!). `deleteNote` is also relevant for this check. Therefore, this cleanup has to be done in a separate method which will be called from these events.

The method itself is very simple. Just the following SQL-query has to be executed: `DELETE FROM categories WHERE id NOT IN (SELECT category FROM notes)`

korelstar:
But for performance reasons, it could be a good option to execute this only once after synchronization. I think, this is the best solution.

app/src/main/java/it/niedermann/owncloud/notes/persistence/NotesDatabase.java
Isaac-Graham: I vote for the second approach, too. But I am thinking about whether renaming the columns of table CATEGORIES is needed. I think that prepending the table name is affordable. What's your opinion?

Isaac-Graham:
> I did a quick review on the code and did some comments.
> 
> I think one thing is missing (at least I didn't found it): what's about removing obsolete categories? If no note is assigned to a category anymore, it should be deleted.

Another question, how to define the obsolete categories? If there is a category with all notes `LOCAL_DELETED`, should this category be removed? Or only those category without any notes should be removed?

Isaac-Graham:
> > For removing obsolete categories, what about doing it in `setCategory(...)`? I think the only one possibility to trigger "obsolete categories" is that the note's category is set to another one. So that we can query in db to find out whether there is such a category without any note and remove it.
> 
> That's not sufficient. There are updates of the category column at least also in `updateNoteAndSync` and `updateNote` (can be called during synchronisazion!). `deleteNote` is also relevant for this check. Therefore, this cleanup has to be done in a separate method which will be called from these events.
> 
> The method itself is very simple. Just the following SQL-query has to be executed: `DELETE FROM categories WHERE id NOT IN (SELECT category FROM notes)`

From the SQL-query you gave, I guess that only those categories without any notes should be removed? In another, if there is a category with all notes `LOCAL_DELETED`, then it should be reserved in the database, right?

Isaac-Graham:
The requests are done, please review it again.

app/src/main/java/it/niedermann/owncloud/notes/persistence/NotesDatabase.java
```diff
@@ -769,4 +802,59 @@ private static void validateAccountId(long accountId) {
...
+    @WorkerThread
+    private Integer getCategoryIdByTitle(long accountId, @NonNull String categoryTitle) {
+        validateAccountId(accountId);
+        SQLiteDatabase db = getReadableDatabase();
+        Cursor cursor = db.query(
+                table_category,
+                new String[]{key_category_id},
+                key_category_title + " = ? AND " + key_category_account_id + " = ? ",
+                new String[]{categoryTitle, String.valueOf(accountId)},
+                key_category_id,
```
korelstar, line 828: Why group by `key_category_id`? Shouldn't this be `null`?

app/src/main/java/it/niedermann/owncloud/notes/persistence/NotesDatabase.java
```diff
@@ -769,4 +802,59 @@ private static void validateAccountId(long accountId) {
...
+                null,
+                key_category_id);
```
korelstar, line 830: Why order by `key_category_id`? Shouldn't this be `null`?

Isaac-Graham:
Thank you for your praise.  Hope we can help for you.

Isaac-Graham:
The ew requests will be done tomorrow since I am going to sleep🥱

app/src/main/java/it/niedermann/owncloud/notes/persistence/NotesDatabase.java
Isaac-Graham: Why is this be null? The column I get is only `key_category_id`. I do not think that this method will crash.

app/src/main/java/it/niedermann/owncloud/notes/persistence/NotesDatabase.java
Isaac-Graham: Why is this be null? The column I get is only `key_category_id`. I do not think that this method will crash.

Isaac-Graham:
I am not sure why you say "getCategoryIdByTitle()" should not be grouped/ordered by `key_category_id`. The only column I select is `key_category_id` and I think it is right.

korelstar:
You are selecting one column and a single row is expected. So why do you want to group and sort this single row? I don't understand your motivation, here. I want to avoid that the DBMS is occupied with things that we don't need.

app/src/main/java/it/niedermann/owncloud/notes/persistence/NotesDatabase.java
korelstar: It's not about crashing, just about performance.

Isaac-Graham:
Oh yes, you are right. I get what you mean.

Isaac-Graham:
Done

app/src/test/java/it/niedermann/owncloud/notes/android/fragment/SearchableBaseNoteFragmentTest.java
```diff
@@ -0,0 +1,106 @@
...
+
+    @Test
+    public void testCountOccurrencesFixed() {
+        try {
+            Method method = SearchableBaseNoteFragment.class.getDeclaredMethod("countOccurrences", String.class, String.class);
+            method.setAccessible(true);
+
+            for (int count = 0; count <= 15; ++count) {
+                StringBuilder sb = new StringBuilder("Mike Chester Wang");
+                Random rand = new Random();
```
stefan-niedermann, line 23: I am not happy with this test.

If it fails, we won't even be able to reproduce it, since we don't have the test parameters in the fail output as far as i can see.

I would prefer to use a fixed value as test parameter (build one with a loop maybe), or at least add all test parameters to fail output.

app/src/main/java/it/niedermann/owncloud/notes/persistence/NotesDatabase.java
```diff
@@ -427,20 +433,41 @@ public void toggleFavorite(SingleSignOnAccount ssoAccount, @NonNull DBNote note,
...
+        values.put(key_category, id);
         db.update(table_notes, values, key_id + " = ?", new String[]{String.valueOf(note.getId())});
+        removeEmptyCategory(note.getAccountId());
         if (callback != null) {
             serverSyncHelper.addCallbackPush(ssoAccount, callback);
         }
         serverSyncHelper.scheduleSync(ssoAccount, true);
     }
 
+    private long addCategory(@NonNull String title, long accountId) {
```
stefan-niedermann, line 462: Switch arguments for consistency

app/src/test/java/it/niedermann/owncloud/notes/android/fragment/SearchableBaseNoteFragmentTest.java
Isaac-Graham: @WingsUpete, @Hui-Ouyang16. Could you please help me to fix this?

app/src/main/java/it/niedermann/owncloud/notes/persistence/NotesDatabase.java
Isaac-Graham: Done

app/src/test/java/it/niedermann/owncloud/notes/android/fragment/SearchableBaseNoteFragmentTest.java
Isaac-Graham: Wait. What do you mean about "not be able to reproduce it"? Do you mean the random?

Isaac-Graham: I guess the reason for "not be able to reproduce" which you said is about the random. In fact, this variable is never used in the test file (I forgot to remove it). I have removed this variable and check again to make sure that there is no useless statement.

app/src/main/java/it/niedermann/owncloud/notes/persistence/AbstractNotesDatabase.java
```diff
@@ -108,10 +115,19 @@ private void createAccountTable(@NonNull SQLiteDatabase db) {
...
+                key_text_color + " VARCHAR(6) NOT NULL DEFAULT '0082C9', " +
+                "FOREIGN KEY(" + key_id + ") REFERENCES " + table_category + "(" + key_category_account_id + "));");
         createAccountIndexes(db);
     }
 
+    private void createCategoryTable(@NonNull SQLiteDatabase db) {
+        db.execSQL("CREATE TABLE " + table_category + "(" +
+                key_category_id + " INTEGER PRIMARY KEY AUTOINCREMENT, " +
+                key_category_account_id + " INTEGER, " +
+                key_category_title + " TEXT )");
```
stefan-niedermann, line 127: Do you guys think a combined `UNIQUE` constraint over accountId and title would make sense?

@korelstar @Isaac-Graham

Isaac-Graham: It is useless but it also can be done. Each time I want to insert a new row into the category. I will check whether there is such a row whose title is the same as what I want to insert (See getCategoryIdByTitle). If there is such a row, I will not insert a row but return the category ID I search. Otherwise, I will insert a new row and return the category ID inserted just now. So that  I think adding a `UNIQUE` contraint is useless.

stefan-niedermann: Yeah, the current implementation of the business level will prevent it, but i think the db should catch as many pitfalls as possible. What can be done at database level should be done there, in this case additionally i think.

It will help us to keep a clean state even if we refactor the business lebel

korelstar: I agree with @stefan-niedermann

PR #791, opened 2020-04-15
Support multiple api levels

stefan-niedermann:
This PR is the groundwork for #786 

It allows to support multiple API levels.

- [x] Store supported API levels in the database
- [x] Update supported API levels periodically and on Pull 2 Refresh
- [x] [Update supported API levels after each sync](https://github.com/stefan-niedermann/nextcloud-notes/issues/786#issuecomment-613039652)
- [x] [Request supported API levels before adding an account](https://github.com/stefan-niedermann/nextcloud-notes/issues/786#issuecomment-613044301)
- [x] Find highest supported API level with fallback to `0.2`

multi-api-level-support

PR #828, opened 2020-05-14
Linking project overview dashboard to README

alexkarezin:
Adding a link to the README header. The linked dashboard describes overall repository structure. The goal is to help new developers explore project and understand architecture/dependencies. Direct link: https://sourcespy.com/github/stefanniedermannnextcloudnotes/

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@ An android client for [Nextcloud Notes App](https://github.com/nextcloud/notes/)
 [![GitHub issues](https://img.shields.io/github/issues/stefan-niedermann/nextcloud-notes.svg)](https://github.com/stefan-niedermann/nextcloud-notes/issues)
 [![GitHub stars](https://img.shields.io/github/stars/stefan-niedermann/nextcloud-notes.svg)](https://github.com/stefan-niedermann/nextcloud-notes/stargazers)
 [![License: GPL v3](https://img.shields.io/badge/License-GPL%20v3-blue.svg)](https://www.gnu.org/licenses/gpl-3.0)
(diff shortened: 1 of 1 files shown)


stefan-niedermann:
Thank you for the contribution. While this is an interesting idea and a nice project, i will not merge this for several reasons:

- The benefit is not yet *that* big. A usual IDE will also be able to create those kind of overview, most people will use Android Studio to contribute to this app
- I think a better contribution would be a structural documentation that is tailored to this project
- If one is interested, a direct communication and explanation is always better than just linking to a generic class diagram
- I will happily explain everyone who asks the architecture of this project
- If one really is looking for a generic generated diagram, one will use google and find the service (s)he likes. No need to link this here.
- The UX of the linked app is.... meh. On a smartphone it is unusable (And yes, that's where is opened it the first time and my first impression was just: meh.) and also on Desktop it lags a good UX in my opinion.

I really like the idea and if it is more powerful and has a better UX, i will reconsider adding a link, but not now, sorry.

alexkarezin:
@stefan-niedermann Thank you very much for taking time to explain the reasoning. I'll go back to the drawing board. One parting question if I may - you mentioned that you would explain architecture of the project to those who are interested. What would it look like and what would you use to illustrate that? Thank you again!

stefan-niedermann:
I think a central element of an architectural overview is the **dataflow**.

Sources (like system events or user interaction), how and where the data gets transformed and the destination of the data (e.g. network calls, data sent to other apps, data displayed on the screen, ...).

One example of a recent PR of this app: https://github.com/stefan-niedermann/nextcloud-notes/pull/791 or a better example from another app where i explained the architecture of push notifications and how they work together: https://github.com/stefan-niedermann/nextcloud-deck/issues/208#issuecomment-609008692

The dataflow is very specific to an app though, therefore i am not sure whether something like this can be generated generically at all.

Issue #786, opened 2020-04-13
Support API v1.0


stefan-niedermann:
1. Add new field `apiVersion` to `Accounts`-table
2. Add `CapabilitiesWorker` which fetches the capabilities
3. If `capabilities` contain `notes`-entry, update `apiVersion` from step 1
4. If `account.apiVersion` is `null`, keep the current behavior
5. If `account.apiVersion` is not `null`, make use of the new features.

I think this will need some changes on the `NotesClient` - we will need to separate the business logic from the technical "connecting to SSO"-part.

korelstar:
Please note, that every response (from Notes version 3.3) will send the HTTP header `X-Notes-API-Versions` which will give the same result like Capabilities API. Therefore, you can update `account.apiVersion` on every request without the need to make an extra request to the Capabilities API.

stefan-niedermann:
Yes i noticed that and i will make use of it, but i will need the capabilities-request also for theming-support.

korelstar:
... and it's definitely required for the first request after adding an account ...

PR #833, opened 2020-05-20
#786 Support API v1.0

stefan-niedermann:
- [x] Make sure excerpt generation is still happening even if title does not match content
- [x] Test with an old Notes app version
  I can confirm that it is working well with an old version (`3.2.0 / API 0.2`) as well as with the current version (`3.5.0 / API 1.1`) of the server app. Unfortunately i haven't been able to test the server upgrade process from an old to the new API version because [i completely fucked up my instance while trying](https://github.com/nextcloud/notes/issues/534). I *assume* that it should work without any problems, but maybe @korelstar has more experience with creating and upgrading Notes server instances and can even cover this case.


stefan-niedermann:
Saw it, thank you very much for the review!

Issue #832, opened 2020-05-20
Move note list widget preferences to database


stefan-niedermann:
Currently all information for note list widgets are stored in `SharedPreferences` with a concatenated key. They should be moved to the database for a better survey.

Depends on: https://github.com/stefan-niedermann/nextcloud-notes/pull/827 (for `CATEGORY_ID`)

This is an ordered snapshot of the `SharedPreferences` with two note list widgets:

```xml
<?xml version='1.0' encoding='utf-8' standalone='yes' ?>
<map>
    <long   name="NLW_account7" value="1" />
    <int    name="NLW_mode7" value="2" />
    <string name="NLW_darkTheme7">SYSTEM_DEFAULT</string>
    <string name="NLW_cat7">Filme</string>
    
    <long   name="NLW_account9" value="1" />
    <int    name="NLW_mode9" value="1" />
    <string name="NLW_darkTheme9">SYSTEM_DEFAULT</string>
</map>
```

Based on this, we need two tables which look like this:

## WIDGET_NOTE_LIST

```
INT ID PK
LONG ACCOUNT_ID FK
LONG CATEGORY_ID FK
LONG MODE_ID ENUM
LONG THEME_MODE ENUM
```

PR #837, opened 2020-05-23
#832 Move settings of NoteListWidget to Database

stefan-niedermann:
This one works pretty good already, but one thing seems to be defect now:

If one has multiple note list widgets on the screen they all display the same content (from the first selected category), while the titles are shown correctly. Also the data stored in the database is correct.

grafik

It think the information gets lost at some point between `NoteListWidget` (correct) and `NoteListWidgetFactory` (for each widget instance the same data - incorrect).

@dan0xii :innocent: :innocent: :innocent: :innocent:


stefan-niedermann:
Doh! Seems like i have removed too much. Will test your changes, thank you!

dan0xii:
There's another issue where the category totals are all being displayed as zero in the configuration activity.

-------- Original Message --------

dan0xii:
> There's another issue where the category totals are all being displayed as zero 
> in the configuration activity.

Fixed. Ready to test.

stefan-niedermann:
Super! Thank you very much :) merged.

PR #840, opened 2020-05-24
#603 sorting method

### LINKED ISSUE #603: option 'last modified' sort on categorized note list needed
As a user I am frequently creating notes and categorising them and within each category I could have many notes (20, for example). However I'm only interested in the most recently used notes. In the current setup some of these notes are in the bottom some of these notes in the middle of the list so it's very hard to distinguish them, especially if they have similar naming conventions. so it would be nice if I could sort by last modified so that I always know which ones I'm actually working on. This is the sort view seen in the 'all' category. 

EDIT: to be clear, I'm talking about the Android app. If this is the wrong issue board, please let me know.

WingsUpete:
Fix #603. We add a column in the category table to store the sorting method for this category note list. Then add an additional icon on the navigation menu at the top for users to select from two sorting methods: last modified desc, lexicographical asc. Some tests are added along.


WingsUpete:
We have done some modifications and we think it is good to check again. @stefan-niedermann

PR #872, opened 2020-06-16
Tweak toolbar design

thgoebel:
This PR tweaks the toolbar in two respects:

The reason being to make it not only more consistent with the files app, but also visually nicer (though the new design was a far bigger jump than these tweaks!).

### 1) Align padddings and elevation

with the files app.

The lesser elevation is especially noticeable in light mode. Having less elevation looks more modern, more like what we see in many of Google's apps implementing Material 2.

Secondly, it adds more margin before and after the card - again, the same values that the files app uses. It also sets the margin to the left of the hamburger to the same value as above and below it.
I conciously did not align the top and bottom margins of the cardview with the files app, so that the height can stay at `?attr/actionBarSize`.

This essentially trades off two design points: giving the toolbar more space (not so squeezed to the screen edge) and having symmetries around the hamburger **versus** aligning the hamburger with the star icons in the recycler. I would argue that the overall impression is nevertheless a cleaner one after these changes. (Especially the hamburger squeezed towards the left edge of the card looked weird.)

| Before  | After  |
|---|---|
|   |   |

### 2) Darken colour of toolbar in darkmode

For that, it uses the same colour as the files app. The old (default) cardColor was a little too bright.

### DIFF
--- a/app/src/main/res/layout/activity_notes_list_view.xml
+++ b/app/src/main/res/layout/activity_notes_list_view.xml
@@ -37,12 +37,12 @@
                 android:id="@+id/home_toolbar"
                 android:layout_width="match_parent"
                 android:layout_height="wrap_content"
-                android:layout_marginStart="@dimen/spacer_1x"
-                android:layout_marginTop="@dimen/design_appbar_elevation"
-                android:layout_marginEnd="@dimen/spacer_1x"
-                android:layout_marginBottom="@dimen/design_appbar_elevation"
+                android:layout_marginStart="@dimen/spacer_2x"
+                android:layout_marginTop="@dimen/spacer_1hx"
+                android:layout_marginEnd="@dimen/spacer_2x"
+                android:layout_marginBottom="@dimen/spacer_1hx"
                 app:cardCornerRadius="@dimen/spacer_1x"
-                app:cardElevation="6dp"
+                app:cardElevation="2dp"
                 app:strokeWidth="0dp">
 
                 <LinearLayout
@@ -55,7 +55,7 @@
                         android:layout_width="wrap_content"
                         android:layout_height="wrap_content"
                         android:background="?attr/selectableItemBackgroundBorderless"
-                        android:paddingStart="@dimen/spacer_1x"
+                        android:paddingStart="@dimen/spacer_2x"
                         android:paddingTop="@dimen/spacer_2x"
                         android:paddingEnd="@dimen/spacer_1x"
                         android:paddingBottom="@dimen/spacer_2x"
--- a/app/src/main/res/layout/activity_notes_list_view.xml
+++ b/app/src/main/res/layout/activity_notes_list_view.xml
@@ -41,6 +41,7 @@
                 android:layout_marginTop="@dimen/spacer_1hx"
                 android:layout_marginEnd="@dimen/spacer_2x"
                 android:layout_marginBottom="@dimen/spacer_1hx"
+                app:cardBackgroundColor="@color/appbar"
                 app:cardCornerRadius="@dimen/spacer_1x"
                 app:cardElevation="2dp"
                 app:strokeWidth="0dp">
--- a/app/src/main/res/values-night/colors.xml
+++ b/app/src/main/res/values-night/colors.xml
@@ -18,4 +18,6 @@
     <color name="fg_contrast">#000000</color>
 
     <color name="category_background">@color/defaultBrand</color>
+
+    <color name="appbar">#1e1e1e</color>
 </resources>
--- a/app/src/main/res/values/colors.xml
+++ b/app/src/main/res/values/colors.xml
@@ -29,6 +29,8 @@
     <color name="category_background">@color/bg_normal</color>
     <color name="category_border">@color/defaultBrand</color>
 
+    <color name="appbar">@android:color/white</color>
+
     <!-- Dark Theme -->
     <!-- Defined here until appwidgets can use night/colors -->
     <color name="fg_default_dark_theme">#eeeeee</color>


stefan-niedermann:
Awesome, thank you :)

1. Could you please adjust the spacing of the "Grid view" (can be enabled in the settings)? Simply search for `new GridItemDecoration` in the `NotesListViewActivity` class.
> [...] aligning the hamburger with the star icons in the recycler. [...]

2. This is my only pain point indeed. In fact we are using not only one but **two** alginments
grafik
We could think about giving the stars more padding but this would result in much lost space...

3. I think there is still a reference to the `4dp` in `appbar_elevation_on.xml`, but i think your intention was to only set the `2dp` for the `MaterialCardView` and keep the `4dp` for the search toolbar, right?

stefan-niedermann:
Maybe we should split this PR down into two separate PRs.
- One for the reduced elevation which we can merge immediately
- The other for the spacing of the toolbar so we can take us time to find out a proper solution regarding the alignment?

thgoebel:
What about the cardColor? A third PR? Happy to split them up though.

stefan-niedermann:
No, the cardColor is fine :+1: keep this change together with the elevation change

thgoebel:
What about moving the star to the end like the GMail has it? Move the category label slightly up, put the star below.

stefan-niedermann:
Would cause a similar issue, because we would have no alignment to the avatar on the right side.

We put it on the left in the first place to get the same position of the elements as in the web UI (and Nextcloud has *all* favorite stars on the left side)

thgoebel:
> I think there is still a reference to the 4dp in appbar_elevation_on.xml, but i think your intention was to only set the 2dp for the MaterialCardView and keep the 4dp for the search toolbar, right?

Yes, I only wanted to change the "normal" card, not the open search.

thgoebel:
Where does `@dimen/design_appbar_elevation` get its 4dp from though? It's not in the dimens.xml :thinking:

jancborchardt:
Good stuff @thgoebel! :)

Same as @stefan-niedermann, my only pain point is the now non-aligned hamburger icon _and_ search text.

Gmail also aligns this:
- Hamburger menu with the list entry icons
- Search input text with the list entry text

We should do that as well, and this is also why it’s good to have the favorite icons on the left.

thgoebel:
But then, do we increase the size of the favorite icons? Because currently the hamburger is too close to the edge of the card.

stefan-niedermann:
We could simply add more padding to the favorite stars, as i said.

But then we need to adjust the search toolbar, the multi selection toolbar and pretty much every other toolbar of the app as well i think 😳

stefan-niedermann:
> @dimen/design_appbar_elevation

ssshhht, it's a property from the material design lib which is marked as private :wink:

stefan-niedermann:
Just had a look at the toolbars and none of them looks aligned 100% correctly. Maybe just add some padding-left to the items is the correct answer...? :confused:

Will need to try this....

stefan-niedermann:
Actually Google doesn't even give a shit about the alignment as it looks:

grafik

@jancborchardt can you give us also a screenshot when one item is selected and one screenshot when there is typed anything in the searchbar?

stefan-niedermann:
Another proposal where i just gave a bit more padding to the stars. The right side would be aligned to the search toolbar because there's a nice alignment with categories.

What do you think @jancborchardt and @thgoebel ? Too much wasted space?

| 1 | 2 |
| --- | --- |
|  |  |

stefan-niedermann:
Please note that i had to move the "Navigate up"-arrow in all other activities, so that they don't jump around while navigating:

grafik

You should consider this for the Files app, too because there it currently looks like this when opening a folder:

grafik

stefan-niedermann:
I put my changes into https://github.com/stefan-niedermann/nextcloud-notes/pull/874

thgoebel:
Continuing the discussion here to keep it in one thread.
For the grid and edit/view-note screens, I think it's okay to have more padding.

But I see some issues:

- (new) the :star:'s left padding is larger than its right padding. 
- (new) for symmetry of the app bar, the ⠇would need more padding on their right
- (old) the padding around the hamburger is still asymmetrical.

stefan-niedermann:
Yeah, i see the trouble. I don't have a clue though how i can keep some alignment.

Because to be honest: I don't like the very first screenshot where the alignment is so extremely off...

Any idea is welcome 🙂

jancborchardt:
I think your screenshots at https://github.com/stefan-niedermann/nextcloud-notes/pull/872#issuecomment-644973426 and https://github.com/stefan-niedermann/nextcloud-notes/pull/872#issuecomment-644988521 are perfect :+1: 

On the issues:

> (new) the star's left padding is larger than its right padding.
> (new) for symmetry of the app bar, the ⠇would need more padding on their right
> (old) the padding around the hamburger is still asymmetrical.

Don’t see these as issues as not all padding has to be the same. The star belongs more to the note title than the edge of the screen. Similarly with the hamburger menu, it’s not all metrics but also visual perception. The hamburger menu being a bit to the left is fine, otherwise it could be mistaken for just an icon of the "Search in all notes" text.

jancborchardt:
> You should consider this for the Files app, too because there it currently looks like this when opening a folder

Indeed @stefan-niedermann → cc @tobiasKaminsky @AndyScherzinger @jorisbodin on the screenshots at https://github.com/stefan-niedermann/nextcloud-notes/pull/872#issuecomment-644988521

stefan-niedermann:
Okay, so let's close this one (thank you though @thgoebel for the initial impulse!)

> (new) for symmetry of the app bar, the ⠇would need more padding on their right

Has been fixed in https://github.com/stefan-niedermann/nextcloud-notes/pull/874

So we have only left

- (new) the star's left padding is larger than its right padding.
- (old) the padding around the hamburger is still asymmetrical.

which is both fine according to @jancborchardt [comment](https://github.com/stefan-niedermann/nextcloud-notes/pull/872#issuecomment-645059094).

Let's get #874 in as it already improves the uniformity with the files app and see if we can improve the layout in further steps.

jancborchardt:
By the way @thgoebel we have a "Notes team" channel on our Nextcloud Talk instance which we use for quicker communication. We can invite you with a guest account if you like? :)

PR #873, opened 2020-06-16
Decrease card elevation and add background color

thgoebel:
Replaces the easy parts of #872.

Curious thing with the files app though - it defines `app:background="@color/appbar"` in `toolbar_standard.xml` but I don't think it actually uses that when build, because `com.google.android.material.card.MaterialCardView` has no attribute `background`. Will need to raise that there...


stefan-niedermann:
Thank you!

PR #874, opened 2020-06-16
Tweak toolbar design based on #872 by @thgoebel

stefan-niedermann:
1. Adds padding to the sides of the card toolbar
2. Adjusts the "navigate up" icon of all activities so they are always at the same position while navigating

Idea by @thgoebel in #872 

| 1 | 2 | 3 |
| --- | --- | --- |
|  |  |  |

grafik


stefan-niedermann:
Quiet more complex than expected :smile:

PR #887, opened 2020-06-20
883 move the fav star in grid mode

### LINKED ISSUE #883: Move the fav star in grid mode
(from original issue https://github.com/stefan-niedermann/nextcloud-notes/issues/881)

When using the phone in one handed mode, the bottom right of a tile is the best area to reach. But this is where the fav star is placed. So it is easy to accidentally fav a note instead of opening it.

Demo video:
https://www.youtube.com/watch?v=BvAaakbesmk

Someone may come up with something better, but this would be one way to solve it:
Move the star icon into the note view.

So let me take the screenshot from https://github.com/stefan-niedermann/nextcloud-notes/issues/882 again and edit it a little. If you fix https://github.com/stefan-niedermann/nextcloud-notes/issues/882, there is enough place in the title bar when in a note to place another button. So it may look like this:

| 1 | 2 |
| --- | --- |
| Screenshot_20200620-111229_flaw2_innote_fixed | Screenshot_20200620-115013_flaw2-grid-fixed |

Now you can remove the star icon from the grid view.

stefan-niedermann:
Fix #883 

(Alternatively keep the category directly next to the star...?)

grafik

### DIFF
--- a/app/src/main/res/layout/item_notes_list_note_item_grid.xml
+++ b/app/src/main/res/layout/item_notes_list_note_item_grid.xml
@@ -52,27 +52,6 @@
             android:orientation="horizontal">
 
             <FrameLayout
-                android:layout_width="0dp"
-                android:layout_height="wrap_content"
-                android:layout_weight="1">
-
-                <com.google.android.material.chip.Chip
-                    android:id="@+id/noteCategory"
-                    android:layout_width="wrap_content"
-                    android:layout_height="wrap_content"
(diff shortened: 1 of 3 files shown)


Bleuzen:
Don't think this is a good idea, see my comment here:
https://github.com/stefan-niedermann/nextcloud-notes/issues/883#issuecomment-646994738

Main point is:
You are just trading right handed users for left handed users.

PR #875, opened 2020-06-17
Improve navdrawer items design

thgoebel:
On API >=21:
- make design of selected item similar to the files app (not pixel-similar, but similar)
- add touch states :innocent: 

On API 17-20 the height slightly increased as well (needed to make it look good on >= 21), but otherwise the look is preserved.

|  API 17-20 | API >=21  |
|---|---|
|   |   |

PR #888, opened 2020-06-20
Remove padding between navigation items again

thgoebel:
#875 introduced some spacing between the navigation drawer items, so that the touch states would not touch (badumtss) each other.

Turns out that might have been too much - with a lot of categories, that's a lot of wasted space. This removes the upper/lower padding again.


stefan-niedermann:
While i initially thought i might loosen up the navigation drawer, i agree that it's too much.

Issue #949, opened 2020-08-30
Support www links without protocol in preview mode


keunes:
Please use **[GitHub reactions](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/)** 👍 to show that you are affected by the same issue. Please don't comment if you have no relevant information to add!

**Describe the bug**

After adding a link, they get rendered as raw text rather than as a link in the preview mode.

**To Reproduce**
1. Open/edit a note
2. Add a link in MD
3. Open preview
4. See error: no link rendered

**Expected behavior**

Render link

**Screenshots**

![Screenshot_20200830-111823.png](https://user-images.githubusercontent.com/11229646/91655668-897d7400-eab2-11ea-8fd8-ec5cd8af12bf.png)

**Smartphone (please complete the following information):**
 - Nextcloud Notes-Version (android app): 2.16.7
 - F-Droid or Play Store: Play
 - Android-Version: 7.1.2
 - Device: FP2

**Server**
 - Nextcloud version: 19.0.2
 - Nextcloud Notes version (server app): ??

**Stacktrace**
N/A

stefan-niedermann:
Ah, can you try to add `http://` / `https://` in front of the link?

keunes:
Ah, that works. If it would recognise `www.` as the start of a URL as well that'd be great. If not with the current lib, then for the new one :)

stefan-niedermann:
Thanks to @muety this will be shipped in version `2.17.0`.

keunes:
Thanks @muety !

PR #183, opened 2017-03-02
linkify URLs in preview

korelstar:
For the notes preview, I've written a regular expression which adds Markdown links `[]()` to URLs, so that they will be click-able. The regular expression does not apply to URLs in existing links (see explanation in the code comments).

Maybe, the RxMD library will have this feature sometime, too -- but until then, we can use this hack.

Furthermore, some minor code improvements were done.

PR #953, opened 2020-09-03
#949 Support www links without protocol in preview mode

stefan-niedermann:
@korelstar i tried to adapt your changes from https://github.com/stefan-niedermann/nextcloud-notes/pull/183 but it doesn't seem to work. Maybe you can give me a hint about the regex here?


korelstar:
I'm not sure about what you exactly want to do. Can you give an example? I did some tests with the following test note:

```
1 A http://www.example.com E
2 A https://www.example.com/ E
3 A www.example.com E
4 A [Test](www.example.com) E
5 A [Test](https://www.example.com/) E
6 A (Test) E
```

Lines 1, 2 and 5 do work already. I just committed an implementation for 3. However, line 4 is not easy to implement false-safe, because we don't want that line 6 (and similar) become a link.

Another thing: you have to decide if the link should become `http://` or `https://`.

stefan-niedermann:
Thank you for the quick response!

I see the trouble. `https://` should become the added protocol, of course.

Couldn't we make a regex which tests for `(` -`stuff except brackets`-`)`-`[`-`here-we-go`-`]`?

stefan-niedermann:
Fixed in `2.19.0`.

PR #987, opened 2020-11-11
Improved Lists

Schnabulator:
As the Commit says I included some Code which makes Shopping Lists way better.
Unchecked and checked items are seperated by a horinzontal line and unchecked items are automatically crossed out

### DIFF
--- a/app/src/main/java/it/niedermann/owncloud/notes/edit/NotePreviewFragment.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/edit/NotePreviewFragment.java
@@ -147,7 +147,31 @@ public class NotePreviewFragment extends SearchableBaseNoteFragment implements O
                                             lines[lineNumber] = lines[lineNumber].replace(CHECKBOX_CHECKED_STAR, CHECKBOX_UNCHECKED_STAR);
                                         }
 
-                                        changedText = TextUtils.join("\n", lines);
+                                        boolean sort = false;
+                                        String unchecked="", checked="";
+                                        for (String i: getResources().getStringArray(R.array.categories_to_sort) ) {
+                                            if (note.getCategory().equals(i)) {
+                                                sort = true;
+                                                // Arrays.sort(lines);
+                                                for (String l: lines){
+                                                    if(l.startsWith("- [ ]")){
+                                                        String _l = l.replaceAll("~~", "");
+                                                        unchecked += _l +"\n";
+                                                    }else if(l.startsWith("- [x]")){
+                                                        String _l = l.replace("- [x] ", "- [x] ~~")+"~~";
+                                                        checked += _l +"\n";
+                                                    }
+                                                }
+
+                                                break;
+                                            }
+                                        }
+                                        if (sort){
+                                            changedText = unchecked+"\n---\n"+checked;
+                                        }else{
+                                            changedText = TextUtils.join("\n", lines);
+                                        }
+
                                         binding.singleNoteContent.setText(parseCompat(markdownProcessor, changedText));
                                         saveNote(null);
                                     } catch (IndexOutOfBoundsException e) {
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
 
-    <string name="app_name">ANotes</string>
-    <string name="app_name_long">Nextcloud ANotes</string>
+    <string name="app_name">Notes</string>
+    <string name="app_name_long">Nextcloud Notes</string>
     <string name="label_all_notes">All notes</string>
     <string name="label_favorites">Favorites</string>
     <string name="action_create">New note</string>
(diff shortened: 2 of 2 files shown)


stefan-niedermann:
Thank you for the contribution! I will review this PR as soon as i can.

stefan-niedermann:
@Schnabulator sorry for the delay 🙁  - i hopefully will find some time next week to review this PR.

Schnabulator:
No problem, i found a "little" Bug in my contribution.
Whenever I have time i will do a little bugfix. 
I appreciate a little feedback if you want to add the feature to the main repository.

stefan-niedermann:
> I appreciate a little feedback if you want to add the feature to the main repository.

Sure, we just have to make sure that it is still fully compatible with the server side app. Ideally we implement it in a way which only affects the visual representation in the Android Apps preview mode and does not modify the actual content.

First feedback after looking at the code:

- Maybe we can utilize the `TextProcessor`
- There are a few cases not covered yet. Have a look at the `MarkdownUtil`, we should support various checkbox formats like `- [ ]`, `* [ ]` (and variations with lower and upper `x` for checked tasks)

Schnabulator:
I dont know how to use the TextProcessor to change e.g. the visualisation of checked items.
To cover all cases I will implement functions in MarkdownUtil.

For your worries I found a better way and right now I'm develop the code for that.

I try to develop a function that will seperate a note into the parts of just text and lists. I will create Code that will only affect on the part of Notes that have Lists. Hope this will work.

stefan-niedermann:
While i appreciate the efforts to display checked list entries like ~~striked through~~, i do not want to not modify the content of the note.
While i see the use case and understand the problem you are trying to solve, i think this might cause lots of confusion, if the app modifies the order of the notes.
We do not understand the semantics of the content.

## Example 1

```
# Tony has to do this:

- [x] Item A
- [ ] Item B

# Peter has to do this:
- [ ] Item C
- [x] Item D
```

:x: How should we know that those task lists in the note are not connected and they should **not** be reordered like

```
# Tony has to do this:

- [x] Item A
- [x] Item D
- [ ] Item B
- [ ] Item C

# Peter has to do this:
```

## Example 2


```
# The order is important

- [ ] 1st: get a job
- [ ] 2nd: earn some money
- [x] 3rd: ???
- [ ] 4th: Profit
```

:x: Also here we don't know anything about the semantic of the note, and it might be important to keep the order.


```
# The order is important

- [x] 3rd: ???
- [ ] 1st: get a job
- [ ] 2nd: earn some money
- [ ] 4th: Profit
```

## Example 3

```
# Tony has to do this:

- [x] Item A
- [ ] Item B

# Peter has to do this:
- [ ] Item C
- [x] Item D
```

:heavy_check_mark: We can safely mark checked items as ~~striked through~~ without knowing anything about the semantic, but :warning: this **must not** modify the note's content but *only* be in the preview view.

```
# Tony has to do this:

- [x] ~~Item A~~
- [ ] Item B

# Peter has to do this:
- [ ] Item C
- [x] ~~Item D~~
```

stefan-niedermann:
> I dont know how to use the TextProcessor to change e.g. the visualisation of checked items.

See [here](https://github.com/stefan-niedermann/nextcloud-notes/blob/510c1051ff2a8da00aba9f494959a7e3543e5a21/app/src/main/java/it/niedermann/owncloud/notes/edit/NotePreviewFragment.java#L237-L241) for how to use the `TextProcessorChain` to move the replacement and processing logic into it's own class. You can call it something like `StrikeCheckedTasksProcessor` or so.

> For your worries I found a better way and right now I'm develop the code for that.

Looking forward :slightly_smiling_face: 

> I try to develop a function that will seperate a note into the parts of just text and lists. I will create Code that will only affect on the part of Notes that have Lists. Hope this will work.

Feel free to do so, but for the [reasons i mentioned in my comment above](https://github.com/stefan-niedermann/nextcloud-notes/pull/987#issuecomment-735442768) i unfortunately can not accept the Pull Request if it changes the semantic of a notes content.

stefan-niedermann:
Closing because of missing feedback and outdated state. Master uses a new rendering engine in the meantime, the logic would be need to rewritten as a proper plugin.

bakeromso:
Just my 2 cents. I liked this feature in Google Keep very much, when packing my bags or grocery shopping, I found it very convenient to keep a view of only what's left to do.
I don't mean to push you to do any work whatsoever, thought I'd just let you know as a user that for my use case, having this option would be wonderful.

rik-shaw:
> Just my 2 cents. I liked this feature in Google Keep very much, when packing my bags or grocery shopping, I found it very convenient to keep a view of only what's left to do. I don't mean to push you to do any work whatsoever, thought I'd just let you know as a user that for my use case, having this option would be wonderful.

Yes that is my feedback as well: having sortable lists (drag n drop up down) and having "checked items" be struck and moved to the bottom would help the use case of it being a grocery shopping list. But I do understand @stefan-niedermann 's comments about not wanting to edit a "real note" under the hood.

Even in Google Keep if you make a "checkbox list" it will not be part of a normal text note. They are exclusive I think for these reasons (that the checkbox list will do things like reorder and reformat etc).

I see apps like claret that have checklists that act like a block put in the middle of a note. But it gets kind of cluttered. So I think having a separate "checlkist" type of note that is not an open / freeform markdown note would make the best use case for a grocery shopping list, but maybe that is outside the scope of Nextcloud Notes?

Issue #994, opened 2020-11-25
App crashed when i opened my shopping list


gergelypolonkai:
Please use **[GitHub reactions](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/)** 👍 to show that you are affected by the same issue. Please don't comment if you have no relevant information to add!

**Describe the bug**

When i opened my shopping list note (a document with some headings, checkboxed lists, and a bunch of emojis), the app crashed.

**To Reproduce**

I cannot willfully reproduce the issue. It happens *sometimes*.

**Expected behavior**

The app should have opened my note.

**Screenshots**

N/A

**Smartphone (please complete the following information):**
 - Nextcloud Notes-Version (android app): 2.17.1
 - F-Droid or Play Store: F-Hroid
 - Android-Version: 10
 - Device: Huawei P20 Pro

**Server**
 - Nextcloud version: 19. 0.5
 - Nextcloud Notes version (server app): 3.6.4

**Stacktrace**

```
App Version: 2.17.1
App Version Code: 2017001
App Flavor: fdroid

Files App Version Code: 30130190

---

OS Version: 4.14.116(10.0.0.171C432)
OS API Level: 29
Device: HWCLT
Manufacturer: HUAWEI
Model (and Product): CLT-L29 (CLT-L29)

---

java.lang.ArrayIndexOutOfBoundsException: length=244; index=-1
	at java.util.ArrayList.get(ArrayList.java:439)
	at it.niedermann.owncloud.notes.main.items.ItemAdapter.getItem(ItemAdapter.java:202)
	at it.niedermann.owncloud.notes.main.MainActivity.onNoteClick(MainActivity.java:867)
	at it.niedermann.owncloud.notes.main.items.NoteViewHolder.lambda$bind$0$NoteViewHolder(NoteViewHolder.java:53)
	at it.niedermann.owncloud.notes.main.items.-$$Lambda$NoteViewHolder$USpBz4_HPjUNfVR-HTVpNrLSa0I.onClick(Unknown Source:2)
	at android.view.View.performClick(View.java:7192)
	at android.view.View.performClickInternal(View.java:7166)
	at android.view.View.access$3500(View.java:824)
	at android.view.View$PerformClick.run(View.java:27592)
	at android.os.Handler.handleCallback(Handler.java:888)
	at android.os.Handler.dispatchMessage(Handler.java:100)
	at android.os.Looper.loop(Looper.java:213)
[...]

Issue #988, opened 2020-11-14
Crash when adding and removing favorite


juliushaertl:
Please use **[GitHub reactions](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/)** 👍 to show that you are affected by the same issue. Please don't comment if you have no relevant information to add!

**Describe the bug**
The app crashes, additionally there is a rendering bug when the favorite was added, could be related.

**To Reproduce**

In the main notes list, mark a note as favorite and remove it again by tapping the star icon

**Expected behavior**

It just works :wink:

**Screenshots**

**Smartphone (please complete the following information):**
 - Nextcloud Notes-Version (android app): 2.17.1
 - F-Droid or Play Store: 
 - Android-Version: 11
 - Device: Pixel 3a

**Server**
 - Nextcloud version: 
 - Nextcloud Notes version (server app): 

**Stacktrace**

```
App Version: 2.17.1
App Version Code: 2017001
App Flavor: play

Files App Version Code: 30140052

---

OS Version: 4.9.223-g90fdcf28fd1c-ab6846391(6869500)
OS API Level: 30
Device: sargo
Manufacturer: Google
Model (and Product): Pixel 3a (sargo)

---

java.lang.ArrayIndexOutOfBoundsException: length=163; index=-1
	at java.util.ArrayList.get(ArrayList.java:439)
	at it.niedermann.owncloud.notes.main.items.ItemAdapter.getItem(ItemAdapter.java:202)
	at it.niedermann.owncloud.notes.main.MainActivity.onNoteFavoriteClick(MainActivity.java:876)
	at it.niedermann.owncloud.notes.main.items.NoteViewHolder.lambda$bindFavorite$2$NoteViewHolder(NoteViewHolder.java:109)
	at it.niedermann.owncloud.notes.main.items.-$$Lambda$NoteViewHolder$rNxJqnoOB-N2wbk5ji9ov7z7Aec.onClick(Unknown Source:2)
	at android.view.View.performClick(View.java:7448)
	at android.view.View.performClickInternal(View.java:7425)
	at android.view.View.access$3600(View.java:810)
	at android.view.View$PerformClick.run(View.java:28305)
	at android.os.Handler.handleCallback(Handler.java:938)
	at android.os.Handler.dispatchMessage(Handler.java:99)
	at android.os.Looper.loop(Looper.java:223)
	at android.app.ActivityThread.main(ActivityThread.java:7656)
[...]

PR #996, opened 2020-12-16
Bug fixed: Use getLayoutPosition instead of getAdapterPosition. #994, #988

Isaac-Graham:
This pull request is used to fix the bugs of #994 and #988.

**The reason of the bugs**
These bugs are both because of the same reason: `getAdapterPosition()` will return `NO_POSITION`, which is -1, to cause the ArrayIndexOutOfBoundsException.

From the comments of `getAdapterPosition()`, we can know that when the layout under re-render or re-calculate, this method will return -1.
During the two layout life-circle, which is an inconsistency less than 16ms, if `getAdapterPosition()` is called, it will return -1.
This leads to two phenomena

1. ArrayIndexOutOfBoundsException with the position being -1.
2. These tricky issues being difficult to reproduce because it is very hard to call this method in such a short time.

**How to reproduce the bugs**
In order to reproduce this bug, I have two possible plans:

1. Use Monkey or some other tools that can click the screen in a very short time.
2. Force the layout to update and then immediately call `getAdapterPosition()`

For me, I use the second plan to reproduce it.
In order to reproduce it, I update the method `onBindViewHolder(RecyclerView.ViewHolder, int)`  in `it.niedermann.owncloud.notes.main.items.ItemAdapter` and `bind(DBNote, boolean, int, int, CharSequence)` in `it.niedermann.owncloud.notes.main.items.NoteViewHolder` as follows:

```java
public void onBindViewHolder(@NonNull final RecyclerView.ViewHolder holder, int position) {
    switch (getItemViewType(position)) {
        case TYPE_SECTION: {
            ((SectionViewHolder) holder).bind((SectionItem) itemList.get(position));
            break;
        }
        case TYPE_NOTE_WITH_EXCERPT:
        case TYPE_NOTE_WITHOUT_EXCERPT:
        case TYPE_NOTE_ONLY_TITLE: {
            ((NoteViewHolder) holder).bind((DBNote) itemList.get(position), showCategory, mainColor, textColor, searchQuery);
            ((NoteViewHolder) holder).itemView.setOnClickListener((view) -> {
                notifyDataSetChanged();
                noteClickListener.onNoteClick(((NoteViewHolder) holder).getAdapterPosition(), view);
            });
            break;
        }
    }
}
```

```java
public void bind(@NonNull DBNote note, boolean showCategory, int mainColor, int textColor, @Nullable CharSequence searchQuery) {
    itemView.setOnClickListener((view) -> noteClickListener.onNoteClick(getAdapterPosition(), view));
    itemView.setOnLongClickListener((view) -> noteClickListener.onNoteLongClick(getAdapterPosition(), view));
}
```

This modified code does the same thing with before. The only difference is that I call `notifyDataSetChanged()` to force the layout updated before `getAdapterPosition()`.

With this modification, the app crashes each time a note is clicked.

**The difference between getAdapterPosition() and getLayoutPosition()**
These two methods are both new API from Google to replace `getPosition()`. The only difference is that `getAdapterPosition()` will return -1 when the layout is updating, which forces developers to treat such 16ms situation. And `getLayoutPosition()` will return a random position between up-to-date position and not-up-to-date position (the positions before and after update).

This will cause such a problem: If `getLayoutPosition()` is called during this 16ms time, a wrong position may be returned.

However, considering the reality that there are few data changes in a short time, I think it is worthy.

### DIFF
--- a/app/src/main/java/it/niedermann/owncloud/notes/main/items/NoteViewHolder.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/main/items/NoteViewHolder.java
@@ -50,8 +50,8 @@ public abstract class NoteViewHolder extends RecyclerView.ViewHolder {
 
     @CallSuper
     public void bind(@NonNull DBNote note, boolean showCategory, int mainColor, int textColor, @Nullable CharSequence searchQuery) {
-        itemView.setOnClickListener((view) -> noteClickListener.onNoteClick(getAdapterPosition(), view));
-        itemView.setOnLongClickListener((view) -> noteClickListener.onNoteLongClick(getAdapterPosition(), view));
+        itemView.setOnClickListener((view) -> noteClickListener.onNoteClick(getLayoutPosition(), view));
+        itemView.setOnLongClickListener((view) -> noteClickListener.onNoteLongClick(getLayoutPosition(), view));
     }
 
     protected void bindStatus(AppCompatImageView noteStatus, DBStatus status, int mainColor) {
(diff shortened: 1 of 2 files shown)


stefan-niedermann (approved):

stefan-niedermann:
A great Pull Request with a preceise description and documentation! Thank you very much for the contribution ❤️

PR #1003, opened 2020-12-27
Avoid overwriting title for new notes

johahauf:
Simple fix to avoid a given title is overwritten. They only whole which still exists is in case the user explicitly wants to use the "default title" as the note title. However I assume an additional flag would be needed to also detect this, but as the likelihood is so low I decided against it.


stefan-niedermann:
I didn't forget your PR, thank you very much for the contribution, i will soon have a look!

PR #1063, opened 2021-02-11
Remove snapshot repo

Unpublished:
Follow up of #1053


stefan-niedermann:
Uhmm, i was so sure that i removed it 😳

Thanks!

Issue #559, opened 2019-05-27
Single Note Widget should handle check list preview


Ornanovitch:
Hi,

As the single note widget is a preview tool, it perfectly handles the preview of title levels with "#", and bullets with "-" or "*", but it does not handles the preview of checkbox with "[ ]", while the preview in the app handles this. 
(Also, it would be the icing on the cake if a check list preview allowed to check or uncheck items directly on the widget, but I keep this idea for a future request :candy: )

Thx for your great work anyway !

A picture so that you can see what I mean : 
![Screenshot_20190527-130505_Evie](https://user-images.githubusercontent.com/34077099/58416828-2b7a0980-8083-11e9-96de-ce452b899eeb.jpg)

stefan-niedermann:
Short answer: It is not possible because Google / Android doesn't allow any other classes than `TextView` (and some irrelevant others) **including** descendants (like our `RxMarkdownTextView`.

Though i did some experiments with search & replace with checkbox emojis - which fails on older devices and some devices which simply do not allow each emoji.

On top of thoses problems, i didn't find a reliable way to detect if a special emoji is really supported.

There seems to be an `EmojiCompat`-library from android where one could have another look, but this seems to be a bit like an overkill for me.

However, the best idea yet for me seems to **replace the characters with an `ImageSpan`**, though i didn't test this yet. Basically this is like implementing some features of `RxMarkdown` ourself 🙁.

If someone else has another idea, feel free to post it or open a PR.

KopfKrieg:
> Short answer: It is not possible because Google / Android doesn't allow any other classes than `TextView` (and some irrelevant others) **including** descendants (like our `RxMarkdownTextView`.

Quick question: Recently RxMarkdown has been replaced by Markwon, but the issue is still the same, right?

> Though i did some experiments with search & replace with checkbox emojis - which fails on older devices and some devices which simply do not allow each emoji.

The ["ballot box" characters are part of Unicode 1.1 from 1993](https://www.compart.com/en/unicode/search?q=Ballot+Box#characters), and [Android 1.5 ships with Unicode 5.0](https://developer.android.com/guide/topics/resources/internationalization).

To me it seems like replacing the checklist items with the unicode "ballot box" characters should be working on all Android devices.

stefan-niedermann:
> Quick question: Recently RxMarkdown has been replaced by Markwon, but the issue is still the same, right?

[...]

PR #1078, opened 2021-03-07
#559 Use emojis for checkboxes in widgets if possible

stefan-niedermann:
For sure not perfect, but a start for #559 (top: old, bottom: new)

grafik

Only Samsung devices will look like shitty teletubby-stuff because they decided to fuck up the used unicode character and use this instead:

grafik
grafik

That's the reason i will use the ugly :heavy_check_mark: and :x: on those devices :man_shrugging:

### DIFF
--- a/fastlane/metadata/android/en-US/changelogs/3002000.txt
+++ b/fastlane/metadata/android/en-US/changelogs/3002000.txt
@@ -1,0 +1,1 @@
+- ✅ Single note widget displays checkboxes more nicely (#559)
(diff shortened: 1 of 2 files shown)


localjo:
I found this issue because I just moved from a Samsung, where the widget was showing :heavy_check_mark: and :x: for my todo items, to a Pixel where I see the unicode box and x. I was searching because I thought emojis were broken on my phone. I thought it might be related to the fonts installed on my device, thinking maybe they didn't support emoji, but now I see that it is by design. While I understand the explanation in this PR, personally, I prefer the emojis because they catch my attention more and tickle the dopamine receptors in my brain (I'm weird, I know). Anyway, it's not super important to me, but if there were an option that I could set to use emoji instead of the plain boxes, I'd love that. Something to consider, anyway.

stefan-niedermann:
Sorry, i currently don't have any plans for this. The intention of the emoji replacements is to display a graphical representation instead of the `- [ ]` and ✔️ and ❌ simply don't look like checkboxes. Of course you can always use the emojis themself directly instead of checkboxes (loosing the other advantages like toggling them by clicking on it) if you prefer them.

PR #1097, opened 2021-03-19
网格试图 -> 网格视图

donglua:


stefan-niedermann:
- Write in english
- Fill in the issue template.

stefan-niedermann:
Ah, i just noticed that this was a Pull Request.

Translations are (only) made on Transifex: https://www.transifex.com/nextcloud/nextcloud/android-notes/

All changes regarding the translations which i merge directly here will be overwritten at the next synchronization with Transifex.

cc @rakekniven

stefan-niedermann:
Ah, i just noticed that this was a Pull Request.

Translations are (only) made on Transifex: https://www.transifex.com/nextcloud/nextcloud/android-notes/

All changes regarding the translations which i merge directly here will be overwritten at the next synchronization with Transifex.

cc @rakekniven

PR #1120, opened 2021-04-08
convert large pngs to webp

connyduck:
This reduces filesize by about 114kB and saves apk size. I used lossless webp encoding so the files look exactly the same.


stefan-niedermann:
Nice, thank you for the contribution!

Can you tell me something about the support of older Android versions? We need to support API level `21` (Android `5.0`).

stefan-niedermann:
> lossless and transparent WebP images are supported in Android 4.3 (API level 18) and higher

according to https://developer.android.com/studio/write/convert-webp :+1:

Issue #1119, opened 2021-04-08
Enable code minification


connyduck:
I noticed this project has [code minification turned off](https://github.com/stefan-niedermann/nextcloud-notes/blob/e7ec28fe3c1116cff4690f85ac109c3088075773/app/build.gradle#L33). I would like to see it turned on because it looks like the savings in app size could be significant. If you want I can send a pull request.

stefan-niedermann:
Nice finding, i just removed it to test it and the file size is (exactly) the same, though... any idea? Have you tried it?

stefan-niedermann:
I was wrong, the size is indeed smaller :slightly_smiling_face:

stefan-niedermann:
YJust wanted to let you know that i remember the reason and that we will disable the minification again 😆 

The reason is that we no longer are able to get meaningful stacktraces, so deciding whether we want a smaller app bundle or more information when a crash happens....

sampe:

```
b.b.a.a.e.n: b.b.a.a.c.c
	at b.b.a.a.d.c.i(Unknown Source:33)
	at b.b.a.a.d.e.c(Unknown Source:6)
	at it.niedermann.owncloud.notes.persistence.p.e(Unknown Source:8)
	at it.niedermann.owncloud.notes.persistence.i.a(Unknown Source:124)
	at it.niedermann.owncloud.notes.main.MainActivity.C0(Unknown Source:71)
	at it.niedermann.owncloud.notes.main.MainActivity.D0(Unknown Source:0)
	at it.niedermann.owncloud.notes.main.l.run(Unknown Source:4)
	at java.lang.Thread.run(Thread.java:919)
```

In case you know how to keep the full class and method names but still generate a smaller bundle size, a PR will be nice 🙂

falzonv:
Hello Stefan,

I may have a solution for this one!

For my launcher application, I am minifying only the Android libraries and all my own classes and method are kept as-is.
This can be done in <code>proguard-rules.pro</code> by telling that everything under *the.package.name.\** should be kept ([example with mine](https://github.com/falzonv/discreet-launcher/blob/main/app/proguard-rules.pro)).

For Nextcloud Notes, it should be something like this to add in your <code>proguard-rules.pro</code> file:

```
# Shrink and obfuscate only Android libraries
-keep class it.niedermann.owncloud.notes.*
{
    <fields> ;
    <init>() ;
    <methods> ;
}
```

[...]

PR #1036, opened 2021-01-10
Remove RxMarkdown completely

stefan-niedermann:
This PR aims to completely remove the not maintained `RxMarkdown`. It is still used to render the content of remote view for widgets, because the don't support custom spans and `Markwon` heavily uses custom spans.

- [ ] Rendering of nested lists doesn't look nice yet
- [ ] Rendering of block quotes doesn't look nice yet
- [ ] Ordered list have no counter

PR #1124, opened 2021-04-09
enable code shrinking

connyduck:
closes https://github.com/stefan-niedermann/nextcloud-notes/issues/1119

This configuration does not obfuscate the bytecode so stacktraces will still be readable. It does some other useful things like removing unused code.

This reduces the size of the app-fdroid-release.apk from 6.964kB to 5.342kB (-23%). With obfuscation it would be 3.949kB (-43%).

(I understand why somebody would not like to use obfuscation, but fyi Google Play deofuscates stacktraces automatically if you use appbundles and you can also do it manually with the retrace tool if you save the optimization.txt files from release builds)

Btw, thank you for this cool app, I use it a lot. :relaxed:

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -30,8 +30,9 @@ android {
         }
 
         release {
-            minifyEnabled false
-            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+            minifyEnabled true
+            shrinkResources true
+            proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
         }
     }
 
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -1,17 +1,7 @@
-# Add project specific ProGuard rules here.
-# By default, the flags in this file are appended to flags specified
-# in C:\Users\stnieder\AppData\Local\Android\sdk/tools/proguard/proguard-android.txt
-# You can edit the include path and order by changing the proguardFiles
-# directive in build.gradle.
-#
-# For more details, see
-#   http://developer.android.com/guide/developing/tools/proguard.html
+# don't obfuscate names and preserve line numbers so crash reports stay readable
+-dontobfuscate
+-keepattributes SourceFile,LineNumberTable
+-renamesourcefileattribute SourceFile
 
-# Add any project specific keep options here:
-
-# If your project uses WebView with JS, uncomment the following
-# and specify the fully qualified class name to the JavaScript interface
-# class:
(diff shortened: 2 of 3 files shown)


stefan-niedermann:
Thank you for the contribution :rocket:

It looks like the `CustomAppGlideModule` isn't available (or working) anymore, at least the account avatar is not longer shown. Can you confirm this?

I really appreciate all the efforts, but i have doubts that it's worth when getting an unpredictable and unreliable `.apk` file in the end which behaves different and in ways and corners which aren't really on our radar. After all it is downloaded *once* and updates can be scheduled. It's more critical for websites which will load the app on every usage (if no service workers are used at least).

Regarding app bundles: I am aware of the benefits, but my last state was that one needs to put the private signing key to Play Store if one wants to use it and is not able to build *and* sign the bundle locally and publish it then... (that is a no-go of course, why i didn't do more research regarding this topic)

connyduck:
> It looks like the CustomAppGlideModule isn't available (or working) anymore, at least the account avatar is not longer shown. Can you confirm this?

My bad. Fixed.

> I really appreciate all the efforts, but i have doubts that it's worth when getting an unpredictable and unreliable .apk file in the end which behaves different and in ways and corners which aren't really on our radar. 

I maintain several heavily shrinked apps and never had such problems. The key is to test release builds extensively before releasing them, because they may differ from debug builds. But I understand if that is too much overhead for you.

> Regarding app bundles: I am aware of the benefits, but my last state was that one needs to put the private signing key to Play Store if one wants to use it and is not able to build and sign the bundle locally and publish it then... (that is a no-go of course, why i didn't do more research regarding this topic)

I have bad news for you: Google will make App Bundles mandatory at the and of this year.

stefan-niedermann:
> My bad. Fixed.

Thanks 🙂 

> The key is to test release builds extensively before releasing them, because they may differ from debug builds.

I got that, but i don't have the resoueces to completely test every feature before every release. Even with automated tests would beed to run against release builds (this should be possible i guess).

Let me think about it and please don't take it as offense if i don't merge the changes. The trade-off is quite high.

> I have bad news for you: Google will make App Bundles mandatory at the and of this year.

Heard that 🙁  Still hoping for some kind of opt-out...

stefan-niedermann:
Did you forget to push your glide fix? I can't see another commit yet and the files look the same 🤷‍♂️ ...

stefan-niedermann:
Sorry, i consider the risc being too high (at least with the current resources for this project). Something slipping through will be quite dangerous and recognized too late by me.

I really appreciate the effort you put into this and i would like to invite you to join #1036 which will be capable of reducing the final app size, too (by removing a complete dependency), but the minification topic is currently out of my scope.

Thank you & regards

PR #1127, opened 2021-04-11
Adds a menu button to remove all selected items from a note

marius56:
This pr adds an additional button "Remove selected items" to the menu when viewing a note.
When clicked, all selected items (e.g. "- [x] Item1\n", "* [X] Item2\n", ...) are removed from the the markdown string using regular expression and the .replaceAll(..) function.

This would be useful for a to do or shopping list. If you completed multiple tasks, you can remove all of them with one click.

In this commit the string is only set for English and German.

### DIFF
--- a/app/src/main/java/it/niedermann/owncloud/notes/edit/NoteEditFragment.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/edit/NoteEditFragment.java
@@ -13,6 +13,7 @@ import android.util.Log;
 import android.util.TypedValue;
 import android.view.LayoutInflater;
 import android.view.Menu;
+import android.view.MenuItem;
 import android.view.View;
 import android.view.ViewGroup;
 import android.view.WindowManager;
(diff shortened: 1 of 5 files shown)


stefan-niedermann:

Unfortunately i have no plans to integrate this feature directly into the app. The use case is too specific and does not justify the introduction of a new menu item which will be visible for each user in the anyway already too crowded menu.

It is a pity that you have done the work for nothing, i recommend you to discuss future feature requests by opening an issue before doing the actual coding part. Also have a look at our [`Good first issues`](https://github.com/stefan-niedermann/nextcloud-notes/labels/Good%20first%20issue) label - they are usually a good starting point if you want to contribute to the app in general.

Besides this, the PR has also some serious quality issues, naming:
- Translations are entirely done via [Transifex](https://www.transifex.com/nextcloud/nextcloud/android-notes/), as stated in the [README.md](https://github.com/stefan-niedermann/nextcloud-notes/blob/master/README.md#rocket-features). The german translation will vanish and be overwritten in the end
- The complete logic and regular expression is duplicated for no reason instead of putting it into the super class or at least a util function
- The regular expression does not respect the [commonmark specification](https://spec.commonmark.org/0.29/) because it does respect checkboxes in code fences (```). This will cause unpredictable behavior and might even lead in data loss
- The wording is imprecise and confusing. First of all, the items are `checked` and not `selected`, also it isn't clear that they will actually vanish instead of just be hidden

PR #1166, opened 2021-04-23
Confirmation dialog for deleting accounts

AlpAcA0072:
Add a simple version of popover confirmation dialog for deleting accounts.

### DIFF
--- a/app/src/main/java/it/niedermann/owncloud/notes/manageaccounts/ManageAccountsActivity.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/manageaccounts/ManageAccountsActivity.java
@@ -1,9 +1,7 @@
 package it.niedermann.owncloud.notes.manageaccounts;
 
 import android.app.AlertDialog;
-import android.content.DialogInterface;
 import android.os.Bundle;
-import android.view.KeyEvent;
 
 import androidx.annotation.Nullable;
 import androidx.appcompat.app.AppCompatActivity;
@@ -17,7 +15,6 @@ import java.util.List;
 
 import it.niedermann.owncloud.notes.LockedActivity;
 import it.niedermann.owncloud.notes.databinding.ActivityManageAccountsBinding;
-import it.niedermann.owncloud.notes.main.MainActivity;
 import it.niedermann.owncloud.notes.shared.model.LocalAccount;
 import it.niedermann.owncloud.notes.persistence.NotesDatabase;
 
@@ -26,32 +23,6 @@ public class ManageAccountsActivity extends LockedActivity {
     private ActivityManageAccountsBinding binding;
     private ManageAccountAdapter adapter;
     private NotesDatabase db = null;
-
-//    protected void dialog() {
-//
-////        AlertDialog.Builder builder = new AlertDialog.Builder(this);
-////        builder.setTitle("确认");
-////        builder.setMessage("这是一个简单消息框");
-////        builder.setPositiveButton("是", null);
-////        builder.show();
-//        AlertDialog.Builder builder = new AlertDialog.Builder(this);
-//        builder.setMessage("确认退出吗？");
-//
-//        builder.setTitle("提示");
-//
-//        builder.setPositiveButton("确认", (dialog, which) -> dialog.dismiss());
-//
-//        builder.setNegativeButton("取消", (dialog, which) -> dialog.cancel());
-//
-//        builder.create().show();
-//    }
-//
-//    public boolean onKeyDown(int keyCode, KeyEvent event) {
-//        if (keyCode == KeyEvent.KEYCODE_BACK && event.getRepeatCount() == 0) {
-//            dialog();
-//        }
-//        return false;
-//    }
 
     @Override
     public void onCreate(@Nullable Bundle savedInstanceState) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/it/niedermann/owncloud/notes/manageaccounts/ManageAccountsActivity.java
```diff
@@ -37,22 +38,36 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
         List<LocalAccount> localAccounts = db.getAccounts();
 
         adapter = new ManageAccountAdapter((localAccount) -> SingleAccountHelper.setCurrentAccount(getApplicationContext(), localAccount.getAccountName()), (localAccount) -> {
-            db.deleteAccount(localAccount);
-            for (LocalAccount temp : localAccounts) {
-                if (temp.getId() == localAccount.getId()) {
-                    localAccounts.remove(temp);
-                    break;
+            AlertDialog.Builder builder = new AlertDialog.Builder(this);
+            builder.setMessage("delete confirmation");
```
stefan-niedermann, line 42: - Please move this to `strings.xml` file (only english), so it can be translated.
- Please use "Removing `%ACCOUNT_NAME%` will also delete unsynchronized notes irrecoverable." as message, to explain the user what's the matter with removing an account

app/src/main/java/it/niedermann/owncloud/notes/manageaccounts/ManageAccountsActivity.java
```diff
@@ -37,22 +38,36 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
...
+
+            builder.setPositiveButton("confirm", (dialog, which) -> {
```
stefan-niedermann, line 44: - Also move to `strings.xml`
- Please use uppercase for the first letter of button actions
- Use "Remove" instead of "Confirm" as verb so the user is actively deciding to execute a destructive action instead of just confirming a dialog (which he will never read) :wink:

app/src/main/java/it/niedermann/owncloud/notes/manageaccounts/ManageAccountsActivity.java
```diff
@@ -37,22 +38,36 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
         List<LocalAccount> localAccounts = db.getAccounts();
 
         adapter = new ManageAccountAdapter((localAccount) -> SingleAccountHelper.setCurrentAccount(getApplicationContext(), localAccount.getAccountName()), (localAccount) -> {
-            db.deleteAccount(localAccount);
-            for (LocalAccount temp : localAccounts) {
-                if (temp.getId() == localAccount.getId()) {
-                    localAccounts.remove(temp);
-                    break;
+            AlertDialog.Builder builder = new AlertDialog.Builder(this);
```
stefan-niedermann, line 41: We have a [`DeleteDialogAlertBuilder`](https://github.com/stefan-niedermann/nextcloud-notes/blob/master/app/src/main/java/it/niedermann/owncloud/notes/branding/BrandedDeleteAlertDialogBuilder.java) class which colors the primary button red - Please use this one instead of the default `AlterDialog.Builder`.

stefan-niedermann (changes requested):
Thank you very much for your contribution :rocket: 
A few details left which should be fixed before merging (see review comments).

Also please add
```java
.setNeutralButton(android.R.string.cancel, null);
```
in addition to the positive button to give the user an obvious way to cancel the current action (clicking on the backdrop works but is not self explaining).

AlpAcA0072:
> Thank you very much for your contribution 🚀
> A few details left which should be fixed before merging (see review comments).
> 
> Also please add
> 
> ```java
> .setNeutralButton(android.R.string.cancel, null);
> ```
> 
> in addition to the positive button to give the user an obvious way to cancel the current action (clicking on the backdrop works but is not self explaining).

Thank you very much for your feedback! I will modify according to them.

app/src/main/java/it/niedermann/owncloud/notes/manageaccounts/ManageAccountsActivity.java
AlpAcA0072: Hi, I wonder how `%ACCOUNT_NAME%` is automatically converted to the name of account, Because it is shown as itself in my tests.

AlpAcA0072: I‘ve finished with that problem and the latest code now satisfied the above requirements. Please check the code again.
Thanks a lot!

AlpAcA0072:
> Thank you very much for your contribution 🚀
> A few details left which should be fixed before merging (see review comments).
> 
> Also please add
> 
> ```java
> .setNeutralButton(android.R.string.cancel, null);
> ```
> 
> in addition to the positive button to give the user an obvious way to cancel the current action (clicking on the backdrop works but is not self explaining).

I‘ve finished with that problem and the latest code now satisfied the above requirements. Please check the code again.
Thanks a lot!

stefan-niedermann:
> I have noticed that I am required to provide some tests below. May I ask if there are any samples for reference?

No worries. I'll merge this PR tomorrow and do some fine tuning in the aftermaths. Good job so far and thank you for the adjustments 🙂

stefan-niedermann (approved):

stefan-niedermann:
Thanks, based on your work i finished this issue. Due to merge conflicts i wasn't able to merge this PR directly, but [i added you explicitly as author of this change](https://github.com/stefan-niedermann/nextcloud-notes/commit/5018bdc956883c87b46ccfa21fa4614ea3aeb67e) in the changelog. :rocket:

Issue #1087, opened 2021-03-12
Option to see how many characters or/and words the note has


moshpirit:
On Nextcloud website version we can see in "Details" how many words and characters the note has, but this feature isn't on Nextcloud Notes. I'd love to have an option to see this in real-time. I don't think it should be difficult to implement, I imagine it like this (it can be in any other edge of the screen or even disappear when scrolling):

![bitmap](https://user-images.githubusercontent.com/7888062/110955935-675e8b80-834a-11eb-8b74-da45d7a3a305.png)

Cui-Yusong:
Hello, I have an idea.
Is it possible to add a  function at the excerpt of the main page that shows the number of words in the text? In this way, the user can know the approximate word count of the article before browsing.

stefan-niedermann:
Closing in favor of https://github.com/stefan-niedermann/nextcloud-notes/issues/1277 - this is nothing i plan to implement on the edit screen, but while redesigning the note details screen, we can add this information there.

PR #1173, opened 2021-04-24
Add function to see how many characters the note has

Cui-Yusong:
By changing the generateNoteExcerpt method, calculate and save the number of characters of the article when generating excerpt.

### DIFF
--- a/app/src/main/java/it/niedermann/owncloud/notes/shared/util/NoteUtil.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/shared/util/NoteUtil.java
@@ -56,17 +56,21 @@ public class NoteUtil {
     }
 
     /**
-     * Generates an excerpt of a content that does <em>not</em> match the given title
+     * Generates word counts and an excerpt of a content that does <em>not</em> match the given title
      *
      * @param content {@link String}
      * @param title   {@link String} In case the content starts with the title, the excerpt should be generated starting from this point
-     * @return excerpt String
+     * @return word counts,excerpt String
      */
     @NonNull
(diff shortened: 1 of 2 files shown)


stefan-niedermann:
However, this is a quick shot and there are so many issues with this PR that i can not merge it.

1. Our workflow / process is to *first discuss ideas* and *then implement it*. This avoids unnecessary work, frustration when something won't get merged and gives us a common, shared idea about the goal
2. I am not sure yet whether we really want to have this feature. I clutters the UI for everyone who don't care about the count and a separate option will need to be tested and maintained in the future.
3. All strings are hard coded, therefore not translatable.
4. You are talking about a "Word count" but implemented a "Character count". So now we have differences in the feature request, the implementation, the description in the UI and the general understanding
5. The feature request was to see the counter **within** the note, not in the list (where it makes even less sense in my personal opinion). Though, let's discuss this before doing such actions. Maybe a circle like a WhatsApp unread counter at the right makes more sense?
5. Simply removing unit tests without an appropriate reason is inacceptible.

I can imagine some kind of new "Details" view for the title, category, modified date and other meta information, though this needs to be thought well and discuessed with the UX / design team of Nextcloud. We will *not* do any quick shot here.

Cui-Yusong:
Yesterday I under https://github.com/stefan-niedermann/nextcloud-notes/issues/1087 provides an idea about the function of the word counting.I personally feel that this solution is not very difficult, so I have already finished this feature before your reply.
I'm sorry that there are still some problems with this idea. I will code after getting your reply next time.
Sorry to get you some troubles.

stefan-niedermann:
Absolutely no problem :slightly_smiling_face: Also please don't feel offended - looking forward to further contributions in the future :)

Issue #1104, opened 2021-03-24
When creating a note named "2021-03-24 - Example title", the separator dash is removed


falzonv:
Please use **[GitHub reactions](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/)** 👍 to show that you are affected by the same issue. Please don't comment if you have no relevant information to add!

**Describe the bug**

- When creating a note with a filename having the format <code>2021-03-24 - Example note title</code>, the dash (-) between the date and the note title is removed and the file is displayed and saved as <code>2021-03-24 Example note title</code>.
- As long as they are not modified, existing notes keep their correct filename but the separator dash is not displayed anymore. On editing and saving the note, the dash is also removed from the filename.
- Using the renaming feature to add the dash does not work (it is removed again on saving).
- Everything works as expected with the Notes app on the server, this issue with the dash happens only in the Android app.
- This bug appeared after upgrading from v3.1.4 to v3.2.2.

**To Reproduce**
1. Create a new note
2. Give it the title <code>2021-03-24 - Example note title</code>
3. Move back to the list of all notes to save
4. The note is now named <code>2021-03-24 Example note title</code> (without the separator dash)
5. Checking on the server, the file have also been saved without the separator dash

**Expected behavior**

The filename should stay as entered by the user.
In this particular case, the dash between the date and the note title should not be removed.

**Screenshots**

Not applicable (well, it could be but I guess the description is clear enough so let's not waste servers space ;-)).

**Smartphone (please complete the following information):**
 - Nextcloud Notes-Version (android app): 3.2.2 (bug appeared right after upgrade from 3.1.4)
 - F-Droid or Play Store: F-Droid
 - Android-Version: 8.1
 - Device: Motorola G5 Plus

**Server**
 - Nextcloud version: 17.0.10
 - Nextcloud Notes version (server app): 3.6.4

**Stacktrace**

Not applicable.

**Additionnal comments**

Thank you very much for providing this awesome application!

stefan-niedermann:
Well described issue, thanks for that 🙂 

I guess it's happening while we try to remove markdown from the title. Probably it is considered as a list item. Will check this.

falzonv:
Hello,

Browsing the commits before v3.2.2 with your hypothesis in mind, it seems the bug was introduced in 4119a926eaf6113373c87fab4b52a342696f11f7.

- In <code>NoteUtil.java</code>, there was the following change:
```
[...]

PR #1174, opened 2021-04-24
Markdown convert to plain text

Jimmy-7664:
Convert markdown to HTML format, then clean HTML tags to get plain text

### DIFF
--- a/markdown/build.gradle
+++ b/markdown/build.gradle
@@ -61,4 +61,5 @@ dependencies {
     testImplementation 'junit:junit:4.13.2'
     androidTestImplementation 'androidx.test.ext:junit:1.1.2'
     androidTestImplementation 'androidx.test.espresso:espresso-core:3.3.0'
+    implementation group: 'com.youbenzi', name: 'MDTool', version: '1.2.4'
 }
--- a/markdown/src/main/java/it/niedermann/android/markdown/MarkdownUtil.java
+++ b/markdown/src/main/java/it/niedermann/android/markdown/MarkdownUtil.java
@@ -17,6 +17,7 @@ import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.core.text.HtmlCompat;
 
+import com.youbenzi.mdtool.tool.MDTool;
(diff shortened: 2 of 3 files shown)


stefan-niedermann:
Sorry, there is lots of trouble with this PR, so [i fixed the issue in the meantime myself](https://github.com/stefan-niedermann/nextcloud-notes/commit/ac114f21c6ed0df62425b578c1b2a744927b7576).

Just a few notes to what i didn't like with this PR to make more clear why i declined it:
1. Introducing a new dependency with the argument to fix a missing dash in a string is not the way we want to go.
2. The idea behind it (converting it to `HTML` and then extract the text) is good: That's why i have the `TODO` in `MarkdownUtil#removeMarkdown`. The problem is: We already have a conversion library: `Markwon` uses the `Commonmark` library in the background which is the natural and only choice for such an approach. You can check out [this PR](https://github.com/stefan-niedermann/nextcloud-notes/pull/1161/files#diff-cf407cc1617606a477850abf942c5518c03c8e48316c8a4a1e46aa8df0368cec) where we are already doing a similar thing for another section of the app.
3. I appreciate you went the extra step to write unit tests, but just removing the existing ones is not pretty nice - ideally they should not even needed to be changed, otherwise we would need to discuss this first.

In general: Thank you very much for your contribution, i would really like to see further contributions from you. But let us discuss the next time **before** you are doing the heavy work, what the goal is and how it should look like in the end :slightly_smiling_face:

Jimmy-7664:
That is my teammates comments in https://github.com/stefan-niedermann/nextcloud-notes/issues/1104 yesterday, and want to solve this problem.
Seeing Falzonv's reply, I think we can do it in this way.During coding, I came up with the idea of converting Markdown to HTML through dependencies.
And we don't get a response until we're done.
I am really sorry for that.Hoping we can do a better cooperation later on.

stefan-niedermann:
Absolutely no problem! As i said: The idea of using `HTML` as an intermediate to get plain text is good, and it should be continued for a really proper solution. Just use the already available `Commonmark` lib instead of introducing a new dependency :slightly_smiling_face:

PR #1178, opened 2021-04-27
pr for issue-1014

Cui-Yusong:
when catch the UnknownErrorException we know that something went wrong with the connection.

Then catch it and then show a snackbar: "please connect to the network"

### DIFF
--- a/app/src/main/java/it/niedermann/owncloud/notes/main/MainActivity.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/main/MainActivity.java
@@ -650,10 +650,12 @@ public class MainActivity extends LockedActivity implements NoteClickListener, A
                                         coordinatorLayout.post(() -> BrandedSnackbar.make(coordinatorLayout, R.string.account_already_imported, Snackbar.LENGTH_LONG).show());
                                     });
                                 } else {
+                                    //CS304 issue: https://github.com/stefan-niedermann/nextcloud-notes/issues/1014
                                     e.printStackTrace();
                                     runOnUiThread(() -> {
                                         binding.activityNotesListView.progressCircular.setVisibility(GONE);
-                                        ExceptionDialogFragment.newInstance(e).show(getSupportFragmentManager(), ExceptionDialogFragment.class.getSimpleName());
+//                                        ExceptionDialogFragment.newInstance(e).show(getSupportFragmentManager(), ExceptionDialogFragment.class.getSimpleName());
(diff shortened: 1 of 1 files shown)


stefan-niedermann:
No. See #1179 for how i meant it.

Your PR has the following issues:
1. No internationalization but a hard coded string which can not be translated. Again.
2. You are canceling the generic handling for all other types of exceptions while we only want to add one more special case for `UnknownErrorException`s *additionally*.
3. Why `coordinatorLayout.post(() -> {});`? We are already running on the UI thread and no longer on a background thread, because we used `runOnUiThread()` before. I guess you just copied the line above, where we need this to get back on the UI thread?

Sorry, please don't feel offended, but i'd like to give you a few advises:
1. Please learn a bit more about the Java and especially Android basics
2. I propose to focus on only one issue at the same time. I appreciate your tries to contribute to the project, but it is better to do *one* issue at a time instead of five and do it well
3. Please consider the feedback that you get - especially the hard coded strings like `"please connect to the network"` is something that [i already told you](https://github.com/stefan-niedermann/nextcloud-notes/pull/1173#issuecomment-826311406) and belongs to the programming fundamentals - this is no knowledge that is taped to Java or Android or this project at all

stefan-niedermann:
I would like to give you another hint:

If you are looking for a way to contribute, i have some tasks which don't require any knowledge of Android and which are pretty fenced.
`MarkdownUtilTest` contains a few [commented test cases](https://github.com/stefan-niedermann/nextcloud-notes/blob/master/markdown/src/androidTest/java/it/niedermann/android/markdown/MarkdownUtilTest.java#L245-L340). The [`MarkdownUtil`](https://github.com/stefan-niedermann/nextcloud-notes/blob/master/markdown/src/main/java/it/niedermann/android/markdown/MarkdownUtil.java) contains the implementation of those test cases and should be adjusted, so that all currently commented unit tests pass.

Maybe this is something you are interested to pick up? Considering the test coverage is quite high for this, you can't do anything wrong :smile: It's pure Java and this part of the app will not directly improve the editor of the Notes app but will also enhance implicitly other apps which use this editor like the Deck app.

Cui-Yusong:
Thanks for your reply.
Quite sorry to have so many mistakes. As you said, my Android knowledge is really poor. 
Thank you very much for providing a task that does not require a lot of Android knowledge. I will try my best to complete it.
I would like to ask you some details about this task. Do you mean that I need to change the TogglePunctuation and InsertLink functions based on these tests so that they can pass the tests that have not passed before. Is that right?

stefan-niedermann:
Yes :)

Issue #1168, opened 2021-04-23
Enable background synchronization by default


stefan-niedermann:
- Enable background synchronization by default for new users
- Enable background synchronization in a migration for all users which have not set explicitly a state yet
- *Optional*: Replace the options with a toggle switch and only provide `On` and `Off`. Use 15 minutes. Since we only use it on Wi-Fi, this won't consume much traffic and the impacts on the battery life is insignificant

Cui-Yusong:
I would like to ask whether the solution to this problem is as follows:
In the original case, background synchronization is not enabled by default.
The change that needs to be made is to enable background synchronization for new users by default.Then add a UI that provides the ON OFF option instead of options.
Is it correct? Thanks very much.

MasterWanna:
I have tried some methods, but they does not work. I have a question that where can I set default value for the variable of `background synchronization` when it is just installed?

stefan-niedermann:
For existing installation we will need a migration. Increment the database version (`NotesDatabase.java`) and add a new `Migration`. There are already migrations which only manipulate the `SharedPreferences`, you can recognize them often on having a `context` parameter.

MasterWanna:
I find that, and all preset values should be set as `Migration`, and to implement this task, I need to add a new `Migration`, is it right?

MasterWanna:
I found a solution which can achieve this, by editing `defaultValue` in `src\main\res\xml\preferences.xml`
![QQ图片20210502150224](https://user-images.githubusercontent.com/70736568/116805209-734f1880-ab57-11eb-984b-f3db93399eac.png)
Is it a avaliable for this?

stefan-niedermann:
Yes, this will work. Though we will still need a migration, because if an user changed the default value in an old version, we need to respect this and migrate the value (`off` or a time interval value) to the new boolean setting.

However, all of this stuff is implementation details, so please continue discussion in a Pull Request. Issues are a matter of "What", Pull Requests are a matter of "How" to implement. 🙂

MasterWanna:
Thanks, got it!

PR #1191, opened 2021-05-02
Enable background sync

MasterWanna:
This is pull request for issue #1168


app/src/main/java/it/niedermann/owncloud/notes/persistence/migration/Migration_21_22.java
```diff
@@ -0,0 +1,34 @@
...
+    /**
+     * Enabling Set backgroundSync to every 15 minutes, and wifiOnly to true
+     * @param database no use just implement
+     */
+    @Override
+    public void migrate(@NonNull SupportSQLiteDatabase database) {
+        SharedPreferences sharedPreferences = PreferenceManager.getDefaultSharedPreferences(context);
+        SharedPreferences.Editor editor = sharedPreferences.edit();
+        editor.putBoolean("wifiOnly", true);
+        editor.putString("backgroundSync", "15_minutes");
```
stefan-niedermann, line 31: We should not only set the `15_minutes` string, but instead
1. Check if there is a preference
2. If there is a preference, remove it
3. Set the preference again, but this time as a `boolean` (`true`)

We should move away from the `String` value and go for only `on` / `off` in the future.

app/src/main/res/xml/preferences.xml
```diff
@@ -11,10 +11,11 @@
             android:icon="@drawable/ic_network_wifi_grey600_24dp"
             android:key="@string/pref_key_wifi_only"
             android:layout="@layout/item_pref"
-            android:title="@string/settings_wifi_only" />
+            android:title="@string/settings_wifi_only"
+            app:defaultValue="true" />
```
stefan-niedermann, line 15: Why did you change the default value for this setting? It should not be touched for fixing the issue.

app/src/main/res/xml/preferences.xml
```diff
@@ -11,10 +11,11 @@
...
 
         <ListPreference
-            android:defaultValue="@string/pref_value_sync_off"
+            android:defaultValue="@string/pref_value_sync_15_minutes"
```
stefan-niedermann, line 18: default value should not be one of thr strings in the future, but simply `true` - a boolean. Since we used a string in older app versions, we need the migration 🙂

stefan-niedermann (changes requested):
Good start. Do you @MasterWanna or @Cui-Yusong have Telegram or Skype messenger by any chance? (I guess you are classmates), maybe we can exchange a bit easier this way?

app/src/main/java/it/niedermann/owncloud/notes/persistence/migration/Migration_21_22.java
```diff
@@ -0,0 +1,34 @@
...
+
+import android.content.Context;
+import android.content.SharedPreferences;
+
+import androidx.annotation.NonNull;
+import androidx.preference.PreferenceManager;
+import androidx.room.migration.Migration;
+import androidx.sqlite.db.SupportSQLiteDatabase;
+
+// CS304 issue link : https://github.com/stefan-niedermann/nextcloud-notes/issues/1168
```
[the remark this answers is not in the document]
MasterWanna: @Cui-Yusong has one, and I will sign up for one then.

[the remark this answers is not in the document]
Cui-Yusong: Yes, I have a telegram account. How could we contact using telegram?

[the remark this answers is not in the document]
Cui-Yusong: Should we create a group chat?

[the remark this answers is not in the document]
stefan-niedermann: Open `https://t.me/` + `stefan` + <kbd>_</kbd> + `niedermann`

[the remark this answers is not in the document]
Cui-Yusong: got it

app/src/main/java/it/niedermann/owncloud/notes/persistence/migration/Migration_21_22.java
```diff
@@ -0,0 +1,33 @@
...
+     * https://github.com/stefan-niedermann/nextcloud-deck/issues/531
+     */
+    @Override
+    public void migrate(@NonNull SupportSQLiteDatabase database) {
+        SharedPreferences sharedPreferences = PreferenceManager.getDefaultSharedPreferences(context);
+        SharedPreferences.Editor editor = sharedPreferences.edit();
+        if (sharedPreferences.contains("backgroundSync")) {
+            editor.remove("backgroundSync");
+        }
+        editor.putBoolean("backgroundSync", true);
```
stefan-niedermann, line 30: The migration should check the current value:
- If it is set to `off`, then use `false` as new boolean value
- If it does not exist, don't explicitly add it (the `defaultValue` will take care of it)
- If it exists and is any other value than `off`, set the preference explicitly to `true` just like you did it here

stefan-niedermann (changes requested):

app/src/main/java/it/niedermann/owncloud/notes/persistence/migration/Migration_21_22.java
MasterWanna: Yes, it should be considered.

app/src/main/java/it/niedermann/owncloud/notes/persistence/SyncWorker.java
```diff
@@ -44,7 +43,14 @@ public Result doWork() {
...
 
-    public static void update(@NonNull Context context, Boolean backgroundSync) {
+    /**
+     * Set up sync work to enabled every 15 minutes or just disabled
+     * https://github.com/stefan-niedermann/nextcloud-notes/issues/1168
+     * @param context the application
+     * @param backgroundSync the toggle result backgroundSync
+     */
+
+    public static void update(@NonNull Context context, @NonNull Boolean backgroundSync) {
```
stefan-niedermann, line 53: Instead of a `@NonNull Boolean`, use just a `boolean` which is basically the same 😉

stefan-niedermann (changes requested):

app/src/main/java/it/niedermann/owncloud/notes/persistence/SyncWorker.java
MasterWanna: Yes, they are the same. I have a question that are there some special cases causes them to be different?

app/src/main/java/it/niedermann/owncloud/notes/persistence/migration/Migration_21_22.java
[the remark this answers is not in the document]
MasterWanna: Hi, it seems that you did not add me, so I cannot send messages.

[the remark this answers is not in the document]
Cui-Yusong: My telegram account is +86 159-0543-1169 , could you please add me? @stefan-niedermann 
I find some problems when adding you .... Thanks~ 😄

app/src/main/java/it/niedermann/owncloud/notes/persistence/SyncWorker.java
stefan-niedermann: The difference is, that a `@NonNull Boolean` **can** be `null`, it is just not *supposed* to be `null`. IDEs like Android Studio will be able to display warnings if you pass `null` into this method, but it will not fail if you actually do it. If you use a value from a unknown source, the annotation will not be able to ensure that there is no `null` passed, while the application directly will fail early at this moment, because the implicit unboxing will throw an exception in this case.

So for primitive values, if you want to make sure, that there is no `null`, go for the primitive type. For all other object types and `String`, go for the annotation.

Other languages like kotlin distinguish between `Boolean` and `Boolean?` as type and provide null safety on compile time, while the `@NonNull` and `@Nullable` annotations in java will not provider compile time "safety" but will support developers well, when they are used across the application. They add more semantic to the wanted type and make the assumptions of the developer more clear.

MasterWanna: Ok, thanks.

app/src/main/java/it/niedermann/owncloud/notes/persistence/migration/Migration_21_22.java
```diff
@@ -0,0 +1,40 @@
...
+        if (sharedPreferences.contains("backgroundSync")) {
+            if (sharedPreferences.getString("backgroundSync", "on").equals("off")) {
+                editor.remove("backgroundSync");
+                editor.putBoolean("backgroundSync", false);
+            } else {
+                editor.remove("backgroundSync");
+                editor.putBoolean("backgroundSync", true);
+            }
+        }
+        editor.apply();
```
stefan-niedermann, line 38: We need to update the `SyncWorker` to make it aware of the "new" setting:

```java
SyncWorker.update(context, true);
```

I'd say we need to call it with `true` in case the `SharedPreferences`
- did **not** contain a setting, because the default behavior now changed
- **did** contain a setting which was **not** `off`

Otherwise the new default will only be applied on new installations and installations which change the setting after performing the update.

I recommend you to test the behavior with the various combinations (No setting, setting `off`, setting `15_minutes`, setting `6_hours`, etc...) with this approach:

1. Install the Notes app on an Android device and set it up
2. Make a snapshot of the emulator (three dots menu -> Take snapshot)
  grafik
3. Validate the old setting in the device explorer of Android Studio
  grafik
4. Install the branch version
5. Validate the new setting in the device explorer (Make sure to right click -> synchronize first, before validating)
6. Check the Logcat (filter for `SyncWorker`) to verify whether it run or not
7. Repeat for each combination (you can now easily restart from your snapshot, see point 2)

stefan-niedermann (changes requested):

app/src/main/java/it/niedermann/owncloud/notes/persistence/migration/Migration_21_22.java
MasterWanna: Is it that we need to call `SyncWorker.update` during `Migration`, otherwise the settings are not updated?

stefan-niedermann: Yes, during the migration.

Oh, and i just saw, that we also need to call it additionally for new installations. So also call it when importing an account the first time:

https://github.com/stefan-niedermann/nextcloud-notes/blob/0348f0a642df5d1a35008f7adcd95de1f4449daf/app/src/main/java/it/niedermann/owncloud/notes/importaccount/ImportAccountActivity.java#L91 (right before `runOnUiThread(() -> {...`) with `true`.

MasterWanna: I have updated in migration, and in `ImportAccountActivity.java`, I dont know whether my operation is correct or not.

app/src/main/java/it/niedermann/owncloud/notes/persistence/migration/Migration_21_22.java
```diff
@@ -0,0 +1,46 @@
...
+        super(21, 22);
+        this.context = context;
+    }
+
+    @Override
+    public void migrate(@NonNull SupportSQLiteDatabase database) {
+        SharedPreferences sharedPreferences = PreferenceManager.getDefaultSharedPreferences(context);
+        SharedPreferences.Editor editor = sharedPreferences.edit();
+        if (sharedPreferences.contains("backgroundSync")) {
+            if (sharedPreferences.getString("backgroundSync", "on").equals("off")) {
```
stefan-niedermann, line 32: Please use `""` as default value instead of `"on"`. The default value should never be used anyway, because we already check with `.contains(...)` the line before, but we never used the value `"on"` and it will be confusing to read it here in the future.

app/src/main/java/it/niedermann/owncloud/notes/persistence/migration/Migration_21_22.java
```diff
@@ -0,0 +1,46 @@
...
+                editor.remove("backgroundSync");
+                editor.putBoolean("backgroundSync", false);
+                SyncWorker.update(context, false);
```
stefan-niedermann, line 35: No need to explicitly call it with `false` in this case, because if the background sync was explicitly switched to `off`, it is already disabled and the behavior will not change while migrating - just remove this line.

app/src/main/java/it/niedermann/owncloud/notes/preferences/PreferencesFragment.java
```diff
@@ -31,12 +29,18 @@
...
     @Override
     public void onCreate(@Nullable Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
     }
 
+    /**
+     * Turn background sync from from {@link String} values to {@link Boolean} values
+     * https://github.com/stefan-niedermann/nextcloud-notes/issues/1168
+     */
+
```
stefan-niedermann, line 43: Remove empty line

app/src/main/java/it/niedermann/owncloud/notes/preferences/PreferencesFragment.java
```diff
@@ -31,12 +29,18 @@
...
 
     @Override
     public void onCreate(@Nullable Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
     }
 
+    /**
+     * Turn background sync from from {@link String} values to {@link Boolean} values
+     * https://github.com/stefan-niedermann/nextcloud-notes/issues/1168
+     */
```
stefan-niedermann, line 42: While this description might fit to the Migration, it does not provide any value to this method. Please remove it or merge it with the description in the Migration.

stefan-niedermann (changes requested):

app/src/main/java/it/niedermann/owncloud/notes/importaccount/ImportAccountActivity.java
```diff
@@ -90,8 +94,17 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
+                            /**
+                             * Update syncing when adding account
+                             * https://github.com/stefan-niedermann/nextcloud-deck/issues/531
+                             * @param account the account to add
+                             */
                             @Override
                             public void onSuccess(Account account) {
+                                Context context = getApplicationContext();
+                                SharedPreferences sharedPreferences = PreferenceManager.getDefaultSharedPreferences(context);
+                                SyncWorker.update(context, !sharedPreferences.getString("backgroundSync", "on").equals("off"));
```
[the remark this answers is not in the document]
MasterWanna: I forgot about this is different to that in `Migration`, this value type should be different to previous value type.

Cui-Yusong:
Hi, I wonder if there's any issues need to be fixed right now? Here's a member of my team feel like to make further knowledge of nextcloud-notes. Thanks a lot！

MasterWanna:
Hi, is there any more I need to improve, or is is Ok in current version?

stefan-niedermann:
No it looks good codewise. Just give me a few days to test each possible combination to make sure no scenario breaks.

I am sorry for the delay, but i am doing this only in my spare time 🙂

stefan-niedermann (approved):

MasterWanna:
Ok, I got it, thanks!

stefan-niedermann:
Good job! Merged!

Issue #1205, opened 2021-05-06
When creating a note starting with "# Test", the "#" character is included in title/filename


falzonv:
Please use **[GitHub reactions](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/)** 👍 to show that you are affected by the same issue. Please don't comment if you have no relevant information to add!

**Describe the bug**

- When creating a note where the first line is a title like `# Test`, the markdown character `#` is included in the filename.
- The same is observed with `## Test`, the two `#` are present at the beginning of the filename.
- This bug appeared after upgrading to v3.4.1 (not sure of the exact previous version but it was at least v3.2.2).
- It may (or not) be related to the fix of #1104, since this was about markdown in the title and was implemented in v3.4.1.

**To Reproduce**

1. Create a new note
2. Write `# Test` on the first line
3. Move back to the list of all notes to save
4. The note now has the title `# Test`
5. Checking on the server, the file have been saved as `# Test.md` (didn't even know it was possible!)

**Expected behavior**

The title markdown character `#` and the space before text should be ignored for defining the title and filename.
In the example above, the title should be `Test` instead of `# Test` and the filename should be `Test.md` instead of `# Test.md`.

**Screenshots**

Not applicable.

**Smartphone (please complete the following information):**
 - Nextcloud Notes-Version (android app): 3.4.1
 - F-Droid or Play Store: F-Droid
 - Android-Version: 8.1
 - Device: Motorola G5 Plus

**Server**
 - Nextcloud version: 20.0.9
 - Nextcloud Notes version (server app): 4.0.4

**Stacktrace**

Not applicable.

AlpAcA0072:
I've located the problem to the method [generateNoteExcerpt](https://github.com/stefan-niedermann/nextcloud-notes/blob/fa9a59aebaa277b2336e619f1a4c4f47cd3166a2/app/src/main/java/it/niedermann/owncloud/notes/shared/util/NoteUtil.java#L66), it seems that the method [removeMarkdown](https://github.com/stefan-niedermann/nextcloud-notes/blob/fa9a59aebaa277b2336e619f1a4c4f47cd3166a2/markdown/src/main/java/it/niedermann/android/markdown/MarkdownUtil.java#L478) with the functionality of removing the markdown syntax is not well implemented as mentioned in [issue#1104](https://github.com/stefan-niedermann/nextcloud-notes/issues/1104#issuecomment-816892217).
[...]

PR #1210, opened 2021-05-11
A simple version of removing Markdown syntax from title

AlpAcA0072:
A simple version of removing Markdown syntax from title. And I'll push forward to the `markwon` method.

### DIFF
--- a/app/src/main/java/it/niedermann/owncloud/notes/shared/util/NoteUtil.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/shared/util/NoteUtil.java
@@ -9,6 +9,8 @@ import androidx.annotation.Nullable;
 
 import it.niedermann.owncloud.notes.R;
 
+import com.youbenzi.mdtool.tool.MDTool;
+
 import static it.niedermann.android.markdown.MarkdownUtil.removeMarkdown;
 import static it.niedermann.android.markdown.MarkdownUtil.replaceCheckboxesWithEmojis;
 
@@ -65,6 +67,10 @@ public class NoteUtil {
     @NonNull
     public static String generateNoteExcerpt(@NonNull String content, @Nullable String title) {
         content = removeMarkdown(replaceCheckboxesWithEmojis(content.trim()));
+        String html = MDTool.markdown2Html(title);
+        if (title != null) {
+            title = html.replaceAll("(<[^<]*?>)|(<[\\s]*?/[^<]*?>)|(<[^<]*?/[\\s]*?>)", "");
+        }
         if (TextUtils.isEmpty(content)) {
             return "";
         }
--- a/app/src/main/java/it/niedermann/owncloud/notes/shared/util/NoteUtil.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/shared/util/NoteUtil.java
@@ -69,11 +69,9 @@ public class NoteUtil {
             return "";
         }
         if (!TextUtils.isEmpty(title)) {
-            if (title != null) {
-                final String trimmedTitle = removeMarkdown(replaceCheckboxesWithEmojis(title.trim()));
-                if (content.startsWith(trimmedTitle)) {
-                    content = content.substring(trimmedTitle.length());
-                }
+            final String trimmedTitle = removeMarkdown(replaceCheckboxesWithEmojis(title.trim()));
+            if (content.startsWith(trimmedTitle)) {
+                content = content.substring(trimmedTitle.length());
             }
         }
         return truncateString(content.trim(), 200).replace("\n", EXCERPT_LINE_SEPARATOR);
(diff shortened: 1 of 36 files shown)


app/src/main/java/it/niedermann/owncloud/notes/shared/util/NoteUtil.java
```diff
@@ -69,9 +69,12 @@ public static String generateNoteExcerpt(@NonNull String content, @Nullable Stri
             return "";
         }
         if (!TextUtils.isEmpty(title)) {
-            final String trimmedTitle = removeMarkdown(replaceCheckboxesWithEmojis(title.trim()));
-            if (content.startsWith(trimmedTitle)) {
-                content = content.substring(trimmedTitle.length());
+            if (title != null) {
+                title = title.replaceFirst("^# ", "");
```
stefan-niedermann, line 73: Why didn't you replace the `# ` characters in `#removeMarkdown()`?

stefan-niedermann (changes requested):

app/src/main/java/it/niedermann/owncloud/notes/shared/util/NoteUtil.java
```diff
@@ -69,9 +69,11 @@ public static String generateNoteExcerpt(@NonNull String content, @Nullable Stri
             return "";
         }
         if (!TextUtils.isEmpty(title)) {
-            final String trimmedTitle = removeMarkdown(replaceCheckboxesWithEmojis(title.trim()));
-            if (content.startsWith(trimmedTitle)) {
-                content = content.substring(trimmedTitle.length());
+            if (title != null) {
```
stefan-niedermann, line 72: Please remove this `if`, it is already covered by `!TextUtils.isEmpty(title)` - actually you can just revert the changes in `NoteUtil.java`.

markdown/src/main/java/it/niedermann/android/markdown/MarkdownUtil.java
```diff
@@ -50,6 +50,7 @@
     private static final Pattern PATTERN_ORDERED_LIST_ITEM = Pattern.compile("^(\\d+).\\s.+$");
     private static final Pattern PATTERN_ORDERED_LIST_ITEM_EMPTY = Pattern.compile("^(\\d+).\\s$");
     private static final Pattern PATTERN_MARKDOWN_LINK = Pattern.compile("\\[(.+)?]\\(([^ ]+?)?( \"(.+)\")?\\)");
+    private static final Pattern PATTERN_MARKDOWN_TITLE = Pattern.compile("^# ");
```
stefan-niedermann, line 53: I am not that good in Regex, but this looks as it would just handle *exactly one* `#` character, right?
It would be a great opportunity to make it work for 1 - 6 `#` characters, so, replace:
1. `# `
2. `## `
3. `### `
4. `#### `
5. `##### `
6. `###### `

stefan-niedermann (changes requested):

markdown/src/main/java/it/niedermann/android/markdown/MarkdownUtil.java
korelstar: Btw. there is already a regex `PATTERN_HEADINGS` that should do the same (but better). (Why) does this not work anymore?

AlpAcA0072:
Thanks for the advice!
I have completed it according to the above two requirements. Please check it again.

korelstar:
Still: why adding a new regex for the same thing like an existing regex should do? The existing regex worked for a long time, and it is still working in version 3.4.1 (just tested). Are there some special cases for which it doesn't work? We should write some tests for those, then.

stefan-niedermann:
@korelstar it has been reported [here](https://github.com/stefan-niedermann/nextcloud-notes/issues/1205) that

> When creating a note where the first line is a title like # Test, the markdown character # is included in the filename.

So something is apparently off. I didn't have time yet to have a closer look at this topic.

korelstar:
Yeah, but I can't reproduce this. Can you?

korelstar:
Here is already a test for this: https://github.com/stefan-niedermann/nextcloud-notes/blob/v3.4.1/markdown/src/androidTest/java/it/niedermann/android/markdown/MarkdownUtilTest.java#L660

Therefore, I think the problem is somewhere else.

stefan-niedermann:
So @AlpAcA0072 sorry for the lost work, but as @korelstar pointed out, removing headlines already works. 

The issue affects only cases where a note is only one line long and i think it can be fixed [here](https://github.com/stefan-niedermann/nextcloud-notes/issues/1205#issuecomment-839757122) - so closing this PR. Thanks though.

stefan-niedermann:
Fixed in https://github.com/stefan-niedermann/nextcloud-notes/commit/3bf57cd2850962045ec5cd8ff8575850770bf029

AlpAcA0072:
Oh I see, may I ask is there any thing else I can do about Markdown?

stefan-niedermann:
No, currently i am not aware of any Markdown related issues which are left...

Issue #1212, opened 2021-05-11
Enhence the markdown manage process in the removeMarkdown()


Cui-Yusong:
Hello, after reading your discussion on the issue about handling markdown, I would like to do the following with the function remove Markdown().
Instead of the original process, I will use the markdown render as you've mentioned before to utilize the markdown text (without adding any new dependencies of course).
With this kind of process, there would not have issues like handling # in the title or something like that. Do you think it is feasible?
Thanks very much 😄

stefan-niedermann:
Sure, go on 😄

PR #1213, opened 2021-05-12
Enhence the markdown manage process

Cui-Yusong:
During the process, I found some test cases are not quite right, and I fixed them.
And I use the render and parse to deal with the markdown.
Is it right? 😃


markdown/src/main/java/it/niedermann/android/markdown/MarkdownUtil.java
```diff
@@ -437,25 +437,20 @@ public static String getMarkdownLink(@NonNull String text, @NonNull String url)
      * @param s Markdown string
      * @return Plain text string
      */
+    @RequiresApi(api = Build.VERSION_CODES.N)
```
stefan-niedermann: Why this annotation? We need to call it from older Android versions, too.

markdown/src/main/java/it/niedermann/android/markdown/MarkdownUtil.java
```diff
@@ -437,25 +437,20 @@ public static String getMarkdownLink(@NonNull String text, @NonNull String url)
...
     @NonNull
+    // CS304 issue link: https://github.com/stefan-niedermann/nextcloud-notes/issues/1212
```
stefan-niedermann, line 441: Please avoid this kind of links, it will be visible in the commit messages if you tag the issue like
> #1212 Use Renderer to remove markdown

markdown/src/main/java/it/niedermann/android/markdown/MarkdownUtil.java
```diff
@@ -437,25 +437,20 @@ public static String getMarkdownLink(@NonNull String text, @NonNull String url)
...
     public static String removeMarkdown(@Nullable String s) {
         if (s == null)
             return "";
         // TODO maybe we can utilize the markwon renderer?
```
stefan-niedermann, line 445: This `TODO` is obsolete when this PR is merged, so remove it

markdown/src/main/java/it/niedermann/android/markdown/MarkdownUtil.java
```diff
@@ -437,25 +437,20 @@ public static String getMarkdownLink(@NonNull String text, @NonNull String url)
...
-        s = PATTERN_SPACE_1.matcher(s).replaceAll("");
-        s = PATTERN_SPACE_2.matcher(s).replaceAll("");
-        return s;
+        // Create HTML string from Markup
+        String html = renderer.render(parser.parse(replaceCheckboxesWithEmojis(s)));
+        html = html.replaceAll("\n","");
+        // Create Spanned from HTML, with special handling for ordered list items
+        Spanned spanned = Html.fromHtml(html,Html.FROM_HTML_MODE_COMPACT);
+        String ans = spanned.toString();
+        ans = ans.substring(0,ans.length()-1);
```
stefan-niedermann, line 452: What's going on here?

markdown/src/main/java/it/niedermann/android/markdown/MarkdownUtil.java
```diff
@@ -438,24 +438,18 @@ public static String getMarkdownLink(@NonNull String text, @NonNull String url)
...
-        s = PATTERN_HEADING_LINE.matcher(s).replaceAll("");
-        s = PATTERN_EMPHASIS.matcher(s).replaceAll("$2");
-        s = PATTERN_SPACE_1.matcher(s).replaceAll("");
-        s = PATTERN_SPACE_2.matcher(s).replaceAll("");
-        return s;
+        // Create HTML string from Markup
+        String html = renderer.render(parser.parse(replaceCheckboxesWithEmojis(s)));
+        html = html.replaceAll("\n","");
+        // Create Spanned from HTML, with special handling for ordered list items
+        Spanned spanned = Html.fromHtml(html,Html.FROM_HTML_MODE_COMPACT);
```
stefan-niedermann, line 450: Funny thing that you mentioned a "special handling for ordered list items" but don't apply it :smile: I guess copy & paste?
Just pass in `null` as third argument and `new ListTagHandler` as fourth argument to `fromHtml`.

stefan-niedermann, line 450: Use `HtmlCompat` in favor of `Html` to maintain compatibility across various Android versions

markdown/src/main/java/it/niedermann/android/markdown/MarkdownUtil.java
```diff
@@ -438,24 +438,18 @@ public static String getMarkdownLink(@NonNull String text, @NonNull String url)
...
-                    s = s.substring(item.length());
-                }
-            }
-        }
-        s = PATTERN_LISTS.matcher(s).replaceAll("");
-        s = PATTERN_HEADINGS.matcher(s).replaceAll("$1");
-        s = PATTERN_HEADING_LINE.matcher(s).replaceAll("");
-        s = PATTERN_EMPHASIS.matcher(s).replaceAll("$2");
-        s = PATTERN_SPACE_1.matcher(s).replaceAll("");
-        s = PATTERN_SPACE_2.matcher(s).replaceAll("");
```
stefan-niedermann, line 458: I guess all of the `PATTERN_*` constants are now unused, right? Please remove them if they aren't needed any longer.

Cui-Yusong:
I found that there are some checks were not success, but I could run it successfully on my computer... Have you ever face this kind of questions?

stefan-niedermann:
> but I could run it successfully on my computer..

Sure? It looks like the usage of `Html` is causing the troubles because the `import` statement is missing - and i can't see an additional `import` in the diff... I guess using `HtmlCompat` instead which already is imported, should solve the trouble with the CI

Cui-Yusong:
Will, the reason that I didn’t use HtmlCompat is that it has some bugs when dealing with  \n .
And Html is the Android’s and don’t need to import.
If I use the HtmlCompat, then there may be some issues dealing with \n and *. Do you want me to use it?

markdown/src/main/java/it/niedermann/android/markdown/MarkdownUtil.java
```diff
@@ -567,18 +567,20 @@ public static String getMarkdownLink(@NonNull String text, @NonNull String url)
...
         String html = renderer.render(parser.parse(replaceCheckboxesWithEmojis(s)));
-        html = html.replaceAll("\n","");
-        // Create Spanned from HTML, with special handling for ordered list items
-        Spanned spanned = Html.fromHtml(html,Html.FROM_HTML_MODE_COMPACT);
-        String ans = spanned.toString();
-        ans = ans.substring(0,ans.length()-1);
-        return ans;
+        html = html.replaceAll("\n", "");
+        // Convert Spanned from HTML. Using HtmlCompat to convert has some problems, which is that sometimes the \n in the text would double.
+        Spanned spanned = HtmlCompat.fromHtml(html, 0);
```
stefan-niedermann, line 577: Have you tried to pass `HtmlCompat.FROM_HTML_MODE_COMPACT` instead of `0`? Maybe this solves the problem with the doubled `\n`?

Cui-Yusong: After trying HtmlCompat.FROM_HTML_MODE_COMPACT instead of 0, the problem of dobled \n is solved. Thanks very much!
Is there any other questions I need to deal with? 😃

Cui-Yusong:
I've changed to using HtmlCompat and use HtmlCompat.FROM_HTML_MODE_COMPACT instead of 0, which could handle the problem of doubled \n. And all the test cases of removeMarkdown() could be passed. 
Is there any problems I need to deal with? 😄

markdown/src/main/java/it/niedermann/android/markdown/MarkdownUtil.java
stefan-niedermann: I am bit scared of a possible `IndexOutOfBoundsException`... Can we please add the following test cases

```java
assertEquals("", MarkdownUtil.removeMarkdown(null));
assertEquals("", MarkdownUtil.removeMarkdown(""));
```

to ensure that it won't crash?

And please change

```java
if (s == null)
    return "";
```

to

```java
if (TextUtils.isEmpty(s))
    return "";
```

This should directly satisfy the added test case.

We also should go for

```java
s = s.trim();
```

instead of messing around with string lengths and indexes to avoid possible errors in case the library behavior changes unexpectedly.

stefan-niedermann:
Nice, two comments are still unresolved in the Pull Request - fix them and we are ready to merge :slightly_smiling_face:

Cui-Yusong:
I've added the tests and used the methods as you suggested. Do you think it's right? 😃

Cui-Yusong:
By the way, is there any  TODO tasks that could be done? I'm very happy to dealing with them 😄 
Thanks very much~

stefan-niedermann:
> I've [...] used the methods as you suggested

Not exactly: I meant `s = s.trim()` as replacement for `s = s.substring(0, s.length() - 1);`

stefan-niedermann:
And those constants aren't used anymore after your refactoring, delete them.

```java
PATTERN_LISTS
PATTERN_HEADINGS
PATTERN_HEADING_LINE
PATTERN_EMPHASIS
PATTERN_SPACE_1
PATTERN_SPACE_2
```

Cui-Yusong:
OK, change as you wished~ 😄

Issue #1169, opened 2021-04-23
Store category sorting for recent, favorites and uncategorized per account


stefan-niedermann:
Currently only "real" categories are stored per account, while the sorting order for "meta" categories is stored globally.
→ Store category sorting method for `RECENT`, `FAVORITES` and `UNCATEGORIZED` per account.

AlpAcA0072:
My understanding is that defaultly there's three categories named `RECENT`, `FAVORITES` and `UNCATEGORIZED` store the corresponding notes. It is right?

AlpAcA0072:
I'm not sure about the meaning of "real" category and "meta" category, could you please explain it?

stefan-niedermann:
Let me explain:

## Current state

Each note can have a category, like `Music` or any other text. Information to those "real" categories (like the sort order) is stored in the `CategoryOptions` table.
Additionally we have some "meta" categories, like `FAVORITES`, `RECENT` and `UNCATEGORIZED` - you can see them in the sidebar and selecting one of them will only show notes from this "meta" category. The difference is, that the relation between those meta categories and each note is not explicitly stored - For example: Selecting all notes of `FAVORITES` will display all notes (no matter in which real category they are), if they have `FAVORITE` property set to `true`.

One can also sort those meta categories (lexicographical or by modified date), just like the real categories, but the data is **not** stored in the `CategoryOptions` table, because only real categories belong into that. Instead they are stored in `SharedPreferences`. A simple key / value store instead of a database table, because we only have a fixed set of meta categories.

While everything is working fine with real categories, the meta categories are currently not aware of multiple accounts - in our key / value store we only have for example:

| key | value |
| --- | --- |
| `FAVORITES` | modified |
| `RECENT` | modified |
| `UNCATEGORIZED` | lexicographically |

## Goal of this issue

1. Do **not** modify anything regarding real categories - they work as they are supposed to.
2. Make meta categories account aware, target state in our key / value store could look for example like this:

| key | value |
| --- | --- |
| `FAVORITES_1` | modified |
| `RECENT_1` | modified |
| `UNCATEGORIZED_1` | lexicographically |
| `FAVORITES_2` | modified |
| `RECENT_2` | modified |
| `UNCATEGORIZED_2` | lexicographically |

(where `_1` and `_2` are referring to the id of the account

[...]

PR #1218, opened 2021-05-16
Store category sorting for recent, favorites and uncategorized per account

AlpAcA0072:
Store category sorting for recent, favorites and uncategorized per account.

### DIFF
--- a/app/src/main/java/it/niedermann/owncloud/notes/persistence/NotesRepository.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/persistence/NotesRepository.java
@@ -177,6 +177,8 @@ public class NotesRepository {
 
     @WorkerThread
     public void deleteAccount(@NonNull Account account) {
+//        final Context ctx = context.getApplicationContext();
+//        final SharedPreferences sp = PreferenceManager.getDefaultSharedPreferences(ctx);
         try {
             ApiProvider.invalidateAPICache(AccountImporter.getSingleSignOnAccount(context, account.getAccountName()));
         } catch (NextcloudFilesAppAccountNotFoundException e) {
@@ -627,20 +629,19 @@ public class NotesRepository {
             int orderIndex = sortingMethod.getId();
             final String categoryType = selectedCategory.getType().toString();
             final String keyMetaCategory = categoryType.concat(String.valueOf(accountId));
-
             switch (selectedCategory.getType()) {
                 case FAVORITES: {
-                    sp.putInt(ctx.getString(R.string.action_sorting_method) + ' ' + ctx.getString(R.string.label_favorites), orderIndex);
+                    sp.putInt(ctx.getString(R.string.action_sorting_method) + ' ' + ctx.getString(R.string.label_favorites) + accountId, orderIndex);
                     sp.putString(keyMetaCategory, "modified");
                     break;
                 }
                 case UNCATEGORIZED: {
-                    sp.putInt(ctx.getString(R.string.action_sorting_method) + ' ' + ctx.getString(R.string.action_uncategorized), orderIndex);
+                    sp.putInt(ctx.getString(R.string.action_sorting_method) + ' ' + ctx.getString(R.string.action_uncategorized) + accountId, orderIndex);
                     sp.putString(keyMetaCategory, "lexicographically");
                     break;
                 }
                 case RECENT: {
-                    sp.putInt(ctx.getString(R.string.action_sorting_method) + ' ' + ctx.getString(R.string.label_all_notes), orderIndex);
+                    sp.putInt(ctx.getString(R.string.action_sorting_method) + ' ' + ctx.getString(R.string.label_all_notes) + accountId, orderIndex);
                     sp.putString(keyMetaCategory, "modified");
                     break;
                 }
@@ -683,19 +684,19 @@ public class NotesRepository {
     public LiveData<CategorySortingMethod> getCategoryOrder(@NonNull NavigationCategory selectedCategory) {
         final SharedPreferences sp = PreferenceManager.getDefaultSharedPreferences(context);
         String prefKey;
-        final String categoryType = selectedCategory.getType().toString();
+        final String accountId = String.valueOf(selectedCategory.getAccountId());
         switch (selectedCategory.getType()) {
             // TODO make this account specific
             case RECENT: {
-                prefKey = context.getString(R.string.action_sorting_method) + ' ' + context.getString(R.string.label_all_notes + '_') + categoryType;
+                prefKey = context.getString(R.string.action_sorting_method) + ' ' + context.getString(R.string.label_all_notes) + accountId;
                 break;
             }
             case FAVORITES: {
-                prefKey = context.getString(R.string.action_sorting_method) + ' ' + context.getString(R.string.label_favorites + '_') + categoryType;
+                prefKey = context.getString(R.string.action_sorting_method) + ' ' + context.getString(R.string.label_favorites) + accountId;
                 break;
             }
             case UNCATEGORIZED: {
-                prefKey = context.getString(R.string.action_sorting_method) + ' ' + context.getString(R.string.action_uncategorized + '_') + categoryType;
+                prefKey = context.getString(R.string.action_sorting_method) + ' ' + context.getString(R.string.action_uncategorized) + accountId;
                 break;
             }
             case DEFAULT_CATEGORY:
--- a/app/src/main/java/it/niedermann/owncloud/notes/persistence/NotesRepository.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/persistence/NotesRepository.java
@@ -458,23 +458,27 @@ public class NotesRepository {
      * @return changed {@link Note} if differs from database, otherwise the old {@link Note}.
      */
     @WorkerThread
-    public Note updateNoteAndSync(Account localAccount, @NonNull Note oldNote, @Nullable String newContent, @Nullable String newTitle, @Nullable ISyncCallback callback) {
+    public Note updateNoteAndSync(@NonNull Account localAccount, @NonNull Note oldNote, @Nullable String newContent, @Nullable String newTitle, @Nullable ISyncCallback callback) {
         final Note newNote;
+        // Re-read the up to date remoteId from the database because the UI might not have the state after synchronization yet
+        // https://github.com/stefan-niedermann/nextcloud-notes/issues/1198
+        @Nullable
+        final Long remoteId = db.getNoteDao().getRemoteId(oldNote.getId());
         if (newContent == null) {
-            newNote = new Note(oldNote.getId(), oldNote.getRemoteId(), oldNote.getModified(), oldNote.getTitle(), oldNote.getContent(), oldNote.getCategory(), oldNote.getFavorite(), oldNote.getETag(), DBStatus.LOCAL_EDITED, localAccount.getId(), oldNote.getExcerpt(), oldNote.getScrollY());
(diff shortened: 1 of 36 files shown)


AlpAcA0072:
I'm still working on deleting meta category when deleting an account.

AlpAcA0072:
#1169

AlpAcA0072:
`Reading` and `updating` are realized as I said in the relevant issue.
As for `deleting`, I get the `SharedPreferences` of current `context` and uses `sharedPreferences.remove()` to remove the related value by the key.

AlpAcA0072:
There's stiil a problem that I don't know how to test the modified code, so I check the keys mannually to make sure they are the same.
The `NotesRepositoryTest` is passed, however, it seems that the relevant tests contain no `SharedPreferences` or `meta category`.

stefan-niedermann:
Yeah, we don't have that much tests (yet). I am working from time to time on the coverage, but that's a huge load of work, given that we started with 0 tests...

I have added a very basic unit test for the delete method of the `NotesRepository` (see [`master` branch](https://github.com/stefan-niedermann/nextcloud-notes/blob/master/app/src/test/java/it/niedermann/owncloud/notes/persistence/NotesRepositoryTest.java#L154-L163)). You should be able to extend it for your use case after rebasing your fork or merging the latest `master`.

AlpAcA0072:
> Yeah, we don't have that much tests (yet). I am working from time to time on the coverage, but that's a huge load of work, given that we started with 0 tests...
> 
> I have added a very basic unit test for the delete method of the `NotesRepository` (see [`master` branch](https://github.com/stefan-niedermann/nextcloud-notes/blob/master/app/src/test/java/it/niedermann/owncloud/notes/persistence/NotesRepositoryTest.java#L154-L163)). You should be able to extend it for your use case after rebasing your fork or merging the latest `master`.

Thank you very much for your hard work and feedback!
I‘ve finished with the new test `testDeleteAccount` and passed all the tests in `NotesRepositoryTest.java`.

app/src/main/java/it/niedermann/owncloud/notes/persistence/NotesRepository.java
```diff
@@ -608,18 +622,22 @@ public void modifyCategoryOrder(long accountId, @NonNull NavigationCategory sele
...
             final SharedPreferences.Editor sp = PreferenceManager.getDefaultSharedPreferences(ctx).edit();
             int orderIndex = sortingMethod.getId();
-
+            final String categoryType = selectedCategory.getType().toString();
+            final String keyMetaCategory = categoryType.concat(String.valueOf(accountId));
             switch (selectedCategory.getType()) {
                 case FAVORITES: {
-                    sp.putInt(ctx.getString(R.string.action_sorting_method) + ' ' + ctx.getString(R.string.label_favorites), orderIndex);
+                    sp.putInt(ctx.getString(R.string.action_sorting_method) + ' ' + ctx.getString(R.string.label_favorites) + accountId, orderIndex);
+                    sp.putString(keyMetaCategory, "modified");
```
stefan-niedermann, line 630: What are the `sp.putString(...)` values needed for...?

stefan-niedermann (changes requested):
I did an early review, and i guess you are overthinking it a bit, especially in the `modifyCategoryOrder` method... I don't really understand why you additionally call `putString()` and place some hard coded strings in the `SharedPreferences`...?

Also: A migration is missing to migrate the existing `SharedPreferences` to the account-specific ones. You can orientate on `Migration_21_22.java` for that.

app/src/main/java/it/niedermann/owncloud/notes/persistence/NotesRepository.java
AlpAcA0072: Ohh I see, I misunderstanded what you meant before. I would delete them.

app/src/main/java/it/niedermann/owncloud/notes/persistence/NotesRepository.java
```diff
@@ -178,6 +178,20 @@ public void addAccount(@NonNull String url, @NonNull String username, @NonNull S
     public void deleteAccount(@NonNull Account account) {
         try {
             ApiProvider.invalidateAPICache(AccountImporter.getSingleSignOnAccount(context, account.getAccountName()));
+            executor.submit(() -> {
```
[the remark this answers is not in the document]
AlpAcA0072: I see, I've removed the `excecutor.submit()`.

AlpAcA0072:
I’ve modified the code according to the above suggestions. 😄

AlpAcA0072:
I don't know exactly what to do in the migration. Should I refresh the sharedPreferences? Or should I uses the refreshed sharedPreferences to update the database using execSQL?

stefan-niedermann:
You basically need to read all account IDs from the database, migrate sharedPreferences as [described here](https://github.com/stefan-niedermann/nextcloud-notes/issues/1169#issuecomment-833539263) and then remove the old sharedPreferences which were not account specific

AlpAcA0072:
> You basically need to read all account IDs from the database, migrate sharedPreferences as [described here](https://github.com/stefan-niedermann/nextcloud-notes/issues/1169#issuecomment-833539263) and then remove the old sharedPreferences which were not account specific

How can I know the structure of the table of database? Without knowing it I could not make a query.

stefan-niedermann:
> How can I know the structure of the table of database?

We are using the [Room](https://developer.android.com/training/data-storage/room/#java) library as [ORM](https://en.wikipedia.org/wiki/Object%E2%80%93relational_mapping). The entities are defined in [`NotesDatabase`](https://github.com/stefan-niedermann/nextcloud-notes/blob/master/app/src/main/java/it/niedermann/owncloud/notes/persistence/NotesDatabase.java#L40-L46) at the very top, each of the class mentioned there matches one table in the database. Each property of a class matches one column of the corresponding table.

Everything you need is to fetch the set of all account IDs and iterate it to add sharedPrefs for each of them.
A similar query is made in the [Migration_22_23](https://github.com/stefan-niedermann/nextcloud-notes/blob/master/app/src/main/java/it/niedermann/owncloud/notes/persistence/migration/Migration_22_23.java#L49-L59).

Don't forget to increment the [database version](https://github.com/stefan-niedermann/nextcloud-notes/blob/master/app/src/main/java/it/niedermann/owncloud/notes/persistence/NotesDatabase.java#L46), otherwise your migration will not be applied (i forgot it a few times myself and was frustrated because it didn't work :wink: )

AlpAcA0072:
I think I've finished with that. Please check whether there is a problem.
Thanks a lot!😆

AlpAcA0072:
> > How can I know the structure of the table of database?
> 
> We are using the [Room](https://developer.android.com/training/data-storage/room/#java) library as [ORM](https://en.wikipedia.org/wiki/Object%E2%80%93relational_mapping). The entities are defined in [`NotesDatabase`](https://github.com/stefan-niedermann/nextcloud-notes/blob/master/app/src/main/java/it/niedermann/owncloud/notes/persistence/NotesDatabase.java#L40-L46) at the very top, each of the class mentioned there matches one table in the database. Each property of a class matches one column of the corresponding table.
> 
> Everything you need is to fetch the set of all account IDs and iterate it to add sharedPrefs for each of them.
> A similar query is made in the [`Migration_22_23`](https://github.com/stefan-niedermann/nextcloud-notes/blob/master/app/src/main/java/it/niedermann/owncloud/notes/persistence/migration/Migration_22_23.java#L49-L59).
> 
> Don't forget to increment the [database version](https://github.com/stefan-niedermann/nextcloud-notes/blob/master/app/src/main/java/it/niedermann/owncloud/notes/persistence/NotesDatabase.java#L46), otherwise your migration will not be applied (i forgot it a few times myself and was frustrated because it didn't work 😉 )

Do I need to add the modified sharedPreferences to the database? I've finished the [`Migration_23_24.java`](https://github.com/AlpAcA0072/nextcloud-notes/blob/5d8b44be4c0a9dbc91b94caed5ae978d6f499ea6/app/src/main/java/it/niedermann/owncloud/notes/persistence/migration/Migration_23_24.java) and want to know is there anything wrong with my implementation

stefan-niedermann:
> Do I need to add the modified sharedPreferences to the database?

No.

But you are removing the keys too early. This way it will only work for one set up account, not in a multi accoubt scenario.
You must only remove them **after** looping over all accounts.

Also please use always hardcoded strings in Migrations (instead of `getString(R.string.xxx)` because the values might change over the time but need to stay the same in migrations forever 🙂

AlpAcA0072:
> > Do I need to add the modified sharedPreferences to the database?
> 
> No.
> 
> But you are removing the keys too early. This way it will only work for one set up account, not in a multi accoubt scenario.
> You must only remove them **after** looping over all accounts.
> 
> Also please use always hardcoded strings in Migrations (instead of `getString(R.string.xxx)`) because the values might change over the time but need to stay the same in migrations forever 🙂

If I hardcoded the string, for example `"Uncategorized"` to replace the `R.string.action_uncategorized`. For any languages other than English, how to guarantee that this method will work?

AlpAcA0072:
[This version](https://github.com/stefan-niedermann/nextcloud-notes/pull/1218/files) satisfy all the requirements above.😄

stefan-niedermann:
You are absolutely right, what a pity that we use localized strings a keys.

We should move away from them during this migration and i unfortunately can not see any other option then gathering a list of all available translated values (see `strings.xml` files), put them into an array in the migration and iterate them.

As a new key you can use something meaningful like
- `category_sorting_` + `favorite_` + `1`
- `category_sorting_` (prefix to distinguish from other sharedPrefs) + `favorite_` (meta category) + `1` (account id)

You can *either* store those new keys as

```java
public static final String XYZ = "category_sorting_";
```

*or* alternatively in the `values/strings.xml` as

```xml
<string name="XYZ" translatable="false">category_sorting_</string>
```

Actually what you have found here is a bug from my point of view, because if one changes the language of his Android device, the sorting of the meta categories won't work.

Some extra miles to go, but worth it.

AlpAcA0072:
> You are absolutely right, what a pity that we use localized strings a keys.
> 
> We should move away from them during this migration and i unfortunately can not see any other option then gathering a list of all available translated values (see `strings.xml` files), put them into an array in the migration and iterate them.
> 
> As a new key you can use something meaningful like
> 
> * `category_sorting_` + `favorite_` + `1`
> * `category_sorting_` (prefix to distinguish from other sharedPrefs) + `favorite_` (meta category) + `1` (account id)
> 
> You can _either_ store those new keys as
> 
> ```java
> public static final String XYZ = "category_sorting_";
> ```
> 
> _or_ alternatively in the `values/strings.xml` as
> 
> ```
> <string name="XYZ" translatable="false">category_sorting_</string>
> ```
> 
> Actually what you have found here is a bug from my point of view, because if one changes the language of his Android device, the sorting of the meta categories won't work.
> 
> Some extra miles to go, but worth it.

So actually what I need to do is to defined three untranslatable meta categories, and migrate the old one to the new one. Because the newly defined one is staticlly stored in sharedPreferences, we don't have to worry about internationalization.
Am I right?
That is such a wonderful idea!

stefan-niedermann:
Yes, it can be considered being a bug to use translated values as keys (which we currently *do*). Therefore we should replace them with static values while we are migrating them anyway.

AlpAcA0072:
I guess now we don't have to worry about hardcoded strings in Migrations.

AlpAcA0072:
Do I have to resolve the conflicts shown below?

stefan-niedermann:
I did not forget you, i am just quite busy with my work and my private life at the moment - sorrs for the delay.

AlpAcA0072:
Thank you for your attention. I am also very busy these days and I hope the repair of the issue will not be delayed.

Issue #993, opened 2020-11-24
Keyboard being dismissed when app moves to background


larena1:
Please use **[GitHub reactions](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/)** 👍 to show that you are affected by the same issue. Please don't comment if you have no relevant information to add!

**Describe the bug**
This issue has been annoying me for quite a while already. When the app moves to background and there's a note open in edit mode with the keyboard active, it will dismiss the keyboard.

That is super annoying because for instance when I make my shopping list, I use to switch between pdf viewer app and notes app and every time I must tap into the note view to continue editing.

Keyboard should not be dismissed when leaving the app. 

**To Reproduce**

**Expected behavior**

**Screenshots**

**Smartphone (please complete the following information):**
 - Nextcloud Notes-Version (android app): 
 - F-Droid or Play Store: 
 - Android-Version: 
 - Device: 

**Server**
 - Nextcloud version: 
 - Nextcloud Notes version (server app): 

**Stacktrace**

larena1:
Is it possible to fix this bug?

stefan-niedermann:
Sure, you can contribute a Pull Request or wait until someone picks it up 🙂 This kind of questions won't help to fix it quicker, though.

larena1:
What would there be needed? Do you think it's just a matter of saving state in onPause and conditionally calling requestFocus in onResume?

stefan-niedermann:
I really don't know yet, maybe that's all what is needed. Give it a try, if it works i will merge it and give you a little bounty 😉

Drhaal:
Are you still interested in this being fixed ? I can start working on it

stefan-niedermann:
Sure, go ahead, a Pull Request will be very welcome 🙂

larena1:
@Drhaal thanks for looking into this. Unfortunately though at least for me it's not working at all.

Using latest notes version on A10 OOS. It's still working fine in Simplenote

stefan-niedermann:
> latest notes version

unfortunately this information is unfortunately quite useless as "latest" is different on Play Store,, AppGallery, F-Droid and other stores and we released this only a few days ago which means that it probably hasn't arrived yet on every device.

The version that ships this MR of @Drhaal is `3.4.11`. I assume `A10` means `Android 10`, but what means `OOS`?

larena1:
[...]

PR #1255, opened 2021-06-08
Keep keyboard state when puting app to the background

Drhaal:
Solves #993

Had to use some workaround to get the state of the keyboard. 
After some manual testing it seems to work well enough and other big app are also relying on this (ex:Firefox mobile). Android does not seem to provide the required functions for this prior to API 30.

Anyway let me know if there is any issue with this PR I'll gladly fix it (if possible)

### DIFF
--- a/app/src/main/java/it/niedermann/owncloud/notes/edit/NoteEditFragment.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/edit/NoteEditFragment.java
@@ -144,15 +144,7 @@ public class NoteEditFragment extends SearchableBaseNoteFragment {
         binding.editContent.addTextChangedListener(textWatcher);
 
         if(keyboardShown){
-            Toast.makeText(requireContext(),"OPEN",Toast.LENGTH_SHORT).show();
-        }
-    }
-
-    @Override
-    protected void onNoteLoaded(Note note) {
-        super.onNoteLoaded(note);
-        if (TextUtils.isEmpty(note.getContent())) {
-            binding.editContent.post(() -> {
+            binding.editContent.postDelayed(() -> {
                 binding.editContent.requestFocus();
 
                 final InputMethodManager imm = (InputMethodManager) requireContext().getSystemService(Context.INPUT_METHOD_SERVICE);
@@ -161,7 +153,24 @@ public class NoteEditFragment extends SearchableBaseNoteFragment {
                 } else {
                     Log.e(TAG, InputMethodManager.class.getSimpleName() + " is null.");
                 }
-            });
+            },100);
+        }
+    }
+
+    @Override
+    protected void onNoteLoaded(Note note) {
+        super.onNoteLoaded(note);
+        if (TextUtils.isEmpty(note.getContent())) {
+            binding.editContent.postDelayed(() -> {
+               binding.editContent.requestFocus();
+
+                final InputMethodManager imm = (InputMethodManager) requireContext().getSystemService(Context.INPUT_METHOD_SERVICE);
+                if (imm != null) {
+                    imm.showSoftInput(binding.editContent, InputMethodManager.SHOW_IMPLICIT);
+                } else {
+                    Log.e(TAG, InputMethodManager.class.getSimpleName() + " is null.");
+                }
+            },100);
         }
 
         binding.editContent.setMarkdownString(note.getContent());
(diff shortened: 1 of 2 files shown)


app/src/main/java/it/niedermann/owncloud/notes/shared/util/DisplayUtils.java
```diff
@@ -52,4 +56,22 @@ private DisplayUtils() {
...
         return new NavigationItem.CategoryNavigationItem("category:" + counter.getCategory(), counter.getCategory(), counter.getTotalNotes(), icon, counter.getAccountId(), counter.getCategory());
     }
+
+    /**
+     * Android does not provide a way to get keyboard visibility prior to API 30 so we use a workaround
+     *
+     * @param parentView View
+     * @return keyboardVisibility Boolean
+     */
+    public static boolean isSoftKeyboardVisible(View parentView) {
```
stefan-niedermann, line 66: Annotate with `@NonNull`

app/src/main/java/it/niedermann/owncloud/notes/edit/NoteEditFragment.java
```diff
@@ -166,11 +163,28 @@ protected void onNoteLoaded(Note note) {
...
+            binding.editContent.requestFocus();
+
+            final InputMethodManager imm = (InputMethodManager) requireContext().getSystemService(Context.INPUT_METHOD_SERVICE);
+            if (imm != null) {
+                imm.showSoftInput(binding.editContent, InputMethodManager.SHOW_IMPLICIT);
+            } else {
+                Log.e(TAG, InputMethodManager.class.getSimpleName() + " is null.");
+            }
+            //Without a small delay the keyboard does not show reliably
+        }, 100);
```
stefan-niedermann, line 177: In which case does it not work reliable? I didn't experience an issue prior to this PR, and i'd like to avoid a delay - even 100ms will make feeling this just laggy...

app/src/main/java/it/niedermann/owncloud/notes/edit/NoteEditFragment.java
```diff
@@ -166,11 +163,28 @@ protected void onNoteLoaded(Note note) {
...
+    }
+
     @Override
     public void onPause() {
         super.onPause();
         binding.editContent.removeTextChangedListener(textWatcher);
         cancelTimers();
+
+        final View parentView = ((ViewGroup) requireActivity().findViewById(android.R.id.content)).getChildAt(0);
```
stefan-niedermann, line 186: This looks a bit riscy in my opinion. As the feature itself is not that important, we should implement it more defensive, add `null` checks and just fallback to `false` if something unexpected happens here. Better be more verbose, the code has already an expiration date (API level 30).

stefan-niedermann (changes requested):
Hello @Drhaal ,

thank you for the contribution 🚀!

You mentioned multiple times that a proper API is available starting with API level 30. But why didn't you use it then and only use the current implementation as a fallback for older devices?

I am fine with having a dirty hack (and that's definitely what it is 😆)
- as long as the proper API is used on SDK 30+ and
- as long as nothing weird or bad happends in case the dirty hack does fail for some reason.

Regarding the latter: I can imagine scenarios with weird display configurations such as foldables, custom control sizes per vendor or even custom keyboard types which might cause fails.
However, the consequences are okay, since in the wrost case the keyboard state just won't be restored (as today), so 👍

app/src/main/java/it/niedermann/owncloud/notes/edit/NoteEditFragment.java
Drhaal: The problem seems to be device dependent.
On my phone the keyboard would not appear when going to background then foreground and not always when creating a new note.
On an emulator it worked fine with both.

I think it's a race conditions problem but I was not able to find the root cause of this.

Drhaal:
I used only one method for consistency but yeah it doesn't make a lot of sense.
I'll add the two methods as soon as possible.

Drhaal:
Add proper keyboard detection for API 30+

app/src/main/java/it/niedermann/owncloud/notes/edit/NoteEditFragment.java
stefan-niedermann: Yep, i was able to confirm those issues. Let's go with `100ms` until we have a better solution :slightly_smiling_face:

stefan-niedermann:
Merged! Thank you very much for your contribution :partying_face:

Issue #1273, opened 2021-06-22
Outline support? Like the one in MarkdownX


dasfsadfaf:
**Is your feature request related to a problem? Please describe.**
Sometimes a note is very big. A outline view helps a lot.

**Describe the solution you'd like**
A outline view helps a lot.  Like the one in MarkdownX

**Describe alternatives you've considered**
none

renfenghuan:
Another man requested the same feature a long time ago, but without reply.
https://github.com/stefan-niedermann/nextcloud-notes/issues/1060#issue-805973782

renfenghuan:
Well, I'm working on it by myself. Looks like not so hard to implement. Maybe I'll start a pull request soon.

renfenghuan:
Done, see https://github.com/stefan-niedermann/nextcloud-notes/pull/1274

stefan-niedermann:
@ghost why did you close this issue and the PR? We haven't merged it yet?

PR #1274, opened 2021-06-22
Add outline support

dasfsadfaf:
Add a new button called "Outline" in the options in the appbar. When pressed, a dialog will pop up. Press certain line to bring you to that header.

See:

Screenshot_2021-06-22-18-15-11-416_it niedermann owncloud notes dev

### DIFF
--- a/app/src/main/java/it/niedermann/owncloud/notes/edit/BaseNoteFragment.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/edit/BaseNoteFragment.java
@@ -6,9 +6,11 @@ import android.content.Intent;
 import android.content.pm.ShortcutInfo;
 import android.content.pm.ShortcutManager;
 import android.graphics.Color;
+import android.graphics.Outline;
 import android.graphics.drawable.Icon;
 import android.os.Build;
 import android.os.Bundle;
+import android.text.Layout;
 import android.util.Log;
 import android.view.Menu;
 import android.view.MenuInflater;
@@ -42,6 +44,9 @@ import it.niedermann.owncloud.notes.accountpicker.AccountPickerDialogFragment;
 import it.niedermann.owncloud.notes.branding.BrandedFragment;
 import it.niedermann.owncloud.notes.edit.category.CategoryDialogFragment;
 import it.niedermann.owncloud.notes.edit.category.CategoryDialogFragment.CategoryDialogListener;
+import it.niedermann.owncloud.notes.edit.outline.OutlineAdapter;
+import it.niedermann.owncloud.notes.edit.outline.OutlineDialogFragment;
+import it.niedermann.owncloud.notes.edit.outline.OutlineItem;
 import it.niedermann.owncloud.notes.edit.title.EditTitleDialogFragment;
 import it.niedermann.owncloud.notes.edit.title.EditTitleDialogFragment.EditTitleListener;
 import it.niedermann.owncloud.notes.persistence.NotesRepository;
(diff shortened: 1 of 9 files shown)


renfenghuan:
Currently this only works the first time the fragment got presented. But I think it's fine because quickly toggle between preview mode and edit mode will refress the outline data.

stefan-niedermann (changes requested):
Hi @renfenghuan ,

thank you very much for your contribution! Sorry for the late reply, it's quite a big addition and i'll definitely review it as soon as possible.

As a quick review:
1. the `outline.type` should be an `Enum`
2. the generation seems not to respect headlines in code blocks (```). This is a quite hard topic, we had the same issues for checkboxes (and [have been able to solve it](https://github.com/stefan-niedermann/nextcloud-notes/blob/master/markdown/src/main/java/it/niedermann/android/markdown/MarkdownUtil.java#L133)). However, our solution is not very generic, but maybe you can reuse some parts of it?
3. The UI does not really satisfy me :smile:
  - The font sizes should never be smaller than factor `1`.
  - The space between the outline items is too big
  - At *least* reuse the multipliers from the editor. You can find them [here](https://github.com/stefan-niedermann/nextcloud-notes/blob/master/markdown/src/main/java/it/niedermann/android/markdown/markwon/plugins/ThemePlugin.java#L32) (maybe need to extract them to `public static final` variables or so)
  - In my opinion they should be all the same size and portrayed like a table of contents with a unordered list and indention per level...?

renfenghuan:
> 
> 
> Hi @renfenghuan ,
> 
> thank you very much for your contribution! Sorry for the late reply, it's quite a big addition and i'll definitely review it as soon as possible.
> 
> As a quick review:
> 
>     1. the `outline.type` should be an `Enum`
> 
>     2. the generation seems not to respect headlines in code blocks (```). This is a quite hard topic, we had the same issues for checkboxes (and [have been able to solve it](https://github.com/stefan-niedermann/nextcloud-notes/blob/master/markdown/src/main/java/it/niedermann/android/markdown/MarkdownUtil.java#L133)). However, our solution is not very generic, but maybe you can reuse some parts of it?
> 
>     3. The UI does not really satisfy me 😄
> 
> 
>     * The font sizes should never be smaller than factor `1`.
> 
>     * The space between the outline items is too big
> 
>     * At _least_ reuse the multipliers from the editor. You can find them [here](https://github.com/stefan-niedermann/nextcloud-notes/blob/master/markdown/src/main/java/it/niedermann/android/markdown/markwon/plugins/ThemePlugin.java#L32) (maybe need to extract them to `public static final` variables or so)
> 
>     * In my opinion they should be all the same size and portrayed like a table of contents with a unordered list and indention per level...?

Thanks for the reply!

1. I made outline.type String because I think in the future, maybe we can add other things other than "Header", maybe split line or something remarkable
2. Yeah I'll look into it
3.  🤣
   a. I tried. If we use 1.45f, 1.35f, 1.25f, 1.15f, 1.1f, 1.05f, the appearance is too subtle. I prefer my current design, maybe we can lower the variance of the factor down a little bit, 
   b. yeah that can be improved .
   c. same as a
   d. Nope. If you see some apps that I suggested in https://github.com/stefan-niedermann/nextcloud-notes/issues/1273, you'll get a bigger picture. Some app adopt your idea, and others adopt my idea. Both are fine. I'm sorry but if you dislike the general appearance you can take it from here and adjust. I'm fine with it right now. Or you can add a setting to toggle what view people want.

renfenghuan:
this's from One Markdown in ios

image

renfenghuan:
this is from Taio in IOS

image

PR #1278, opened 2021-06-23
#1277 Redesign note details

### LINKED ISSUE #1277: Redesign note details
Currently the options menu of a note is quite cluttered, what makes it hard to distinguish between the various entries and spreads meta information across different menus.

The idea is to design on unified dialog which covers

- Title
- Category
- Favorite
- Share
- Pin to homescreen
- (optionally in the future, because this is a harder task: ) Moving across accounts

We can also add the sometimes requested modified date to this view, since we now have more space in a place which will not clutter the UI with information that might not be interesting for the majority of the users and also get some more space for new content related features like an outline in the options menu.

# :eyes: Mock-Up

[mockup](https://user-images.githubusercontent.com/4741199/123069881-883b8000-d413-11eb-8b22-fe0dfe0f0455.png)

stefan-niedermann:
Fix #1277 

image


korelstar:
I like it, but I'm unsure if "save" is the correct wording. Maybe "OK" is better?

According to the Material Design Guidelines, I would understand this dialog as confirmation dialog, which should have buttons Cancel and OK.
https://material.io/components/dialogs#confirmation-dialog

stefan-niedermann:
I am not yet sure how i want to implement it - either save implicitly, just like the notes itself, or explicitly using a button (as it is at the moment). By the way the favorite state is saved directly, regardless of the buttom you choose - it felt unnatural to only save it after pr

From various people i learned that

Issue #1348, opened 2021-09-15
"&lt;" in a text string


Valdnet:
There is a character `&lt;` in the text string that is misinterpreted by Transifex:
![obraz](https://user-images.githubusercontent.com/47037905/133383988-2990a158-2891-400b-ad63-afc20e3b52cc.png)

https://github.com/stefan-niedermann/nextcloud-notes/blob/f1a88971e2b9ec84be596cb062d97b79fc82e553/app/src/main/res/values/strings.xml#L84

It should be correct for Transifex:

![obraz](https://user-images.githubusercontent.com/47037905/133384026-caf1aec0-18c0-4d53-b027-18de8c5eb303.png)

Translators get an error when translating in Transifex:

![obraz](https://user-images.githubusercontent.com/47037905/133384867-abbe8990-8f6b-47a9-8717-5ed2bb3cc0e9.png)

I will make changes to all text strings where `&lt;`.

Valdnet:
Please make changes for other text strings as well.

This requires more work in Java code. I leave the corrections for developers.
See: [#1349](https://github.com/stefan-niedermann/nextcloud-notes/pull/1349#issuecomment-919764561).

stefan-niedermann:
@Valdnet i removed all `&lt;` as well as all HTML markup from the strings in https://github.com/stefan-niedermann/nextcloud-notes/commit/ff2cab8f5c932ce50480728c008d83d66ce37f5b#diff-5e01f7d37a66e4ca03deefc205d8e7008661cdd0284a05aaba1858e6b7bf9103

> It is very urgent.

I don't doubt that, though those strings have been present since 5 years^^

Valdnet:
> though those strings have been present since 5 years^^

Yes, but after changing this string from PR: #1347, translators have a problem with translating it and saving it in Transifex.

Thanks 😉.

PR #1349, opened 2021-09-15
i18n: Change "&lt;" on "<"

Valdnet:
Issue: #1348.


stefan-niedermann:
This might cause issues because it will be recognized as XML...

I think the better way is to get rid of html in the translations at all (we did this in the [Deck Android app](https://github.com/stefan-niedermann/nextcloud-deck/blob/master/app/src/main/res/values-de/strings.xml) too, but this requires a few changes in the Java code)

Valdnet:
Okay, then we are waiting for this to be fixed because these text strings are causing serious confusion among translators. They cannot be translated correctly. **It is very urgent**.

stefan-niedermann:
Understood. I should be able to fix this within a week :)

PR #1362, opened 2021-10-03
#522 Overflow-menu contains "search" when searchview is open

### LINKED ISSUE #522: Overflow-menu contains "search" when searchview is open 🔎
This is a follow-up to #400, detected by @korelstar :

On my phone, I have some weird behavior regarding the menu items: By default, the search menu item is shown in the ActionBar. If I start a search and type some chars, the search menu item is not shown anymore (which is intended). However, if I use the three-dot-menu then, the search is still active and the search menu item is shown in the pull down menu. After closing and reopening the three-dot-menu again, the search menu item isn't shown anymore. This behavior is reproducible (Android 6.0.1).

vhouriet:
Hey, I removed the condition to collapse the search menu. It should now works as intended, when opening the menu, it now collapse the search menu in both Edit and Preview mode end there is no "search" item in the menu list. I don't know if that breaks anything though.

Original behaviour :
before

New behaviour:
after

Fix #522


stefan-niedermann:
@vhouriet thank you very much for your contribution :rocket: :pray: 

Fix #522

Issue #1379, opened 2021-10-18
Opening note from widget when device credentials is enabled empties the note.


hugalafutro:
Please use **[GitHub reactions](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/)** 👍 to show that you are affected by the same issue. Please don't comment if you have no relevant information to add!

**Describe the bug**

I enabled Device Credentials security in the Notes app to prevent accidentally opening notes by touching homepage widget. However when I touch the widget now and scan my fingerprint - the note opens as blank and the data is gone and I have to roll back the version in Nextcloud Files to get it back.

**To Reproduce**
1. Enable Device Credentials in Security options of the Notes app
2. Exit the app
3. Create a homepage Notes widget
4. Touch any note in the widget
5. Device will ask for its biometric/pattern/pin lock
6. Unlock
7. The note that will load will be empty and the data will be gone

**Expected behavior**
The note should display normally

**Screenshots**

N/A

**Smartphone (please complete the following information):**
 - Nextcloud Notes-Version (android app): 3.4.16
 - F-Droid or Play Store: Play Store
 - Android-Version: 11
 - Device: OnePlus 6T with OxygenOS 11.1.1.1

**Server**
 - Nextcloud version: 22.2.0 
 - Nextcloud Notes version (server app): 4.1.1

**Stacktrace**
No ide what that is. There is nothing coming up in NC server log while this is happening.

PR #1400, opened 2021-11-11
1379 widget credentials

nwatson3:
Possible fix for #1379
I think the bug is being caused by the note's content in NoteEditFragment being overwritten with an empty content string from the TextView when the note gets saved before it loads the content into the TextView, and this save may be being caused by the credentials check.
Not pulling from the view, but from the note object itself before any content has been loaded seems to fix.
Sorry if this is missing something obvious, it's my first time working on this project.


stefan-niedermann:
Dear @nwatson3 thank you very much for your PR 🚀! Please give me some time to review and test the changes 🙂

stefan-niedermann:
Sorry for the delay, i did not forget you, just quite busy at the moment 😕

Issue #1461, opened 2022-02-06
Option to keep the screen on


renekliment:
Please use **[GitHub reactions](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/)** 👍 to show that you are affected by the same issue. Please don't comment if you have no relevant information to add!

**Is your feature request related to a problem? Please describe.**

When viewing / editing some notes I don't interact with the device for a while and I don't want the screen to turn off.
In these cases it is annoying to have to unlock the device quite often.
My example cases:
- I am writing something and I need quite some time to think what to write next.
- I am reading something and I think about it for a while.
- I am viewing a note with instructions for whatever and I keep looking at it periodically. Sometimes I have my hands dirty / have gloves on / whatever and unlocking the device is even more annoying than in the previous cases. 

**Describe the solution you'd like**

It would be awesome if there was an option in the settings to keep the screen on when viewing / editing the notes.

**Describe alternatives you've considered**

I could turn off the system-wide screen sleep altogether, but I don't really want that for other apps so it would impact my device battery negatively.

**Additional context**

None. I really love the app, this is the one UX improvement that would make it perfect for me. :slightly_smiling_face:

stefan-niedermann:
I agree, that's a nice idea - I don't think we'll need an option for that, so we'll simply keep the screen on always if one is watching an actual note (editing or viewing). The default behavior is of course to switch the display off when the system requests to do so in other cases (list view, settings view, formatting view, about view, ...) :)

Fixed in https://github.com/stefan-niedermann/nextcloud-notes/commit/d721cb128d837f8fb15e3d55565439c09fa44ef5 (will be in version `3.4.20`).

PR #1496, opened 2022-04-10
#1461 Add an option of keep screen on

anshangPro:
this pr is connected with issue #1461.
add a flag in the EditNoteActivity.
Every time creating the activity, it will check the flag statement and set the windows' flag FLAG_KEEP_SCREEN_ON.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,5 +6,3 @@
 /.idea/
 *.iml
 /projectFilesBackup/
-/app/release/
-/app/schemas/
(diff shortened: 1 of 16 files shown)


stefan-niedermann:
Hey @anshangPro 

thanks for your contribution, that's a nice idea!

I think we can just implement it without a setting - simply always set the flag (should be a one-liner).

Would you change your PR?

anshangPro:
> 

Hi, @stefan-niedermann 

Thanks, for reviewing.

I think I can't just set the flag without the setter. 

Once you open a new page to edit notes, EditNoteActivity will create a new instance. But the windows flags are connected to a living instance. So instead of just getting the window of activity and changing the flag in the listener, I choose to set a boolean so that when creating the activity, the activity itself can check the boolean and set the flag.

I am still a learner, so I am not sure whether some other better implements exist. If you have some good ideas, I am pleased to implement them.

stefan-niedermann:
Well, too long to explain because it was literally one line :laughing:
Check out https://github.com/stefan-niedermann/nextcloud-notes/commit/d721cb128d837f8fb15e3d55565439c09fa44ef5 for the solution :)

Issue #1519, opened 2022-05-09
Feature Request: Switch to editing on double tap


newhinton:
Please use **[GitHub reactions](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/)** 👍 to show that you are affected by the same issue. Please don't comment if you have no relevant information to add!

**Is your feature request related to a problem? Please describe.**
Usually i want to see a note fully rendered. If i now want to make an edit, i have to switch modes via the menuicon in the topbar. (This is expected and totally fine for markdown apps)

**Describe the solution you'd like**
I suggest a double-tap feature. Usually when the user long presses text the context menu appears. This could be extended with a double tap gesture, which would toggle the mode from viewer to editor. If the cursor also jumps to the tapped position, editing becomes even faster

**Describe alternatives you've considered**
None, the current way works and this suggestion is just an improvement

**Additional context**
No additional context.

teachmain:
I think I can do that. ：）

stefan-niedermann:
I am not sure yet whether I am that happy with this kind of solution. Double tapping usually triggers on Android a "Select word" action and it also interfers with checkboxes.

I think there was a feature request some time ago to make the two views "swipable" - like with tabs, but no "tab headers". Personally I think there are less obstacles from an UX point of view.

stefan-niedermann:
I thought about it and while I understand the requirement for a quicker switch between preview and edit mode, the drawbacks of the suggested solution are too high in my opinion.

> Double tapping usually triggers on Android a "Select word" action and it also interfers with checkboxes.

We can discuss other options. I think we once discussed being able to switch via a swipe gesture like with tabs (without actual tab headers), but double tap is definitely out (and I recommend the same for the server app)

newhinton:
@stefan-niedermann I agree with your conclusion!

One idea i had was to introduce a FAB. Usually that is more easily reachable by a thumb, in comparison to the topbar. It would not introduce ambiguity like the double-tab, but it would take up more space as the topbar. Maybe that would help?

bonbonboi:
I don't recommended this request. It's enough I'm already annoyed by the text selection when I accidentally touch the screen in preview mode while reading the note.

PR #1526, opened 2022-05-30
#1519 Switch to editing on double tap 

teachmain:
I try to change `EditNoteActivity` as less as I can, add double tap gesture to enter edit mode from preview mode. The interval of double tap is const 500ms for now.


stefan-niedermann:
Closing because of https://github.com/stefan-niedermann/nextcloud-notes/issues/1519#issuecomment-1167072504

PR #1534, opened 2022-06-15
i18n: Spelling unification

Valdnet:
Signed-off-by: Valdnet <47037905+Valdnet@users.noreply.github.com>


stefan-niedermann:
Phew, that's a tricky one because these strings are *not* part of the apps UI but used to recognize common categories (defined by the user).

If you write a note, I think it is more common to name thr category `ToDo` than `To-Do` (at least in germany), do you agree?

It was possible to add `To-Do` as an *additional* indicator for this kind of category, but I wouldn't use it as only option...

Valdnet:
If this is difficult to do, we can leave it as it is 😉.

PR #1536, opened 2022-06-25
allow EditNoteActivity to open main application when called from widget

### LINKED ISSUE #1412: Back button in widget-link
Please use **[GitHub reactions](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/)** 👍 to show that you are affected by the same issue. Please don't comment if you have no relevant information to add!

**Is your feature request related to a problem? Please describe.**
I'm always frustrated when i am opening a note from a widget, and then want to switch the note. Instead of opening the main list, the view closes back to the homescreen.

**Describe the solution you'd like**
Redirect the user to the main screen of the app instead of closing

** Alternatives: **
 Replace the 'back-arrow' icon with a close icon, to reduce confusion

newhinton:
This fixes #1412

When we close() the activity, currently we only call finish() after cleanup. This results in the currently visible activity to be removed. However, if this activity is actually the current root of the application, it ends the application altogether.

This is often the case when the app was called from a widget, and hasn't been used for a while. Android cleans note's backstack, and then the EditNoteAcitivity-Instance is actually the root instance. Instead of finishing this, we redirect the app to the MainActivity which is the desired target.


stefan-niedermann:
I think it was a requirement at some point in the past (by @jancborchardt ?) that the app should explicitly **not** show the note list when closing the editor after launching it from a widget. I unfortunately wasn't able to find the related discussion yet ...

newhinton:
Hm yes, i checked my calendar-app and the widget there does the same thing. However, does it make sense?

I can only speak for myself, but if i open a note, it is very likely that i want to edit a different note aswell, especially when i use the back button on the top.

If i want to exit, i would usually use the home button/gesture to exit it completely.

How do we find out what people want?


As a contra-point: K9-Mail opens the mail from the widget, and if the user goes 'back', it opens the main view aswell.

jancborchardt:
Totally agree actually. I have the same problem with the single note widget.

I edit something in single note, want to go to the notes list to edit some other note, using the top left "back" button, but then the app closes instead.

I'd be all for always opening the notes list when pressing the top left back button. :)

newhinton:
@jancborchardt Does this also apply to the same issue in the deck-app? It may be a different use case, so maybe it is different there. (https://github.com/stefan-niedermann/nextcloud-deck/issues/1293)

jancborchardt:
@newhinton I'd say yep let's do the same there. Then it's consistent: Whenever you are in a card and press top left (in-app) back button, you also stay in the app.

newhinton:
Any news on this?
I can fix this in the deck app aswell on the weekend if you want.

stefan-niedermann:
Sure, let's bring this in. Would you mind rebasing the branch? Then I'll merge it 🙂

newhinton:
Done! The automatic merge was possible.

PR #1544, opened 2022-07-05
Add monochrome icon

### LINKED ISSUE #1543: Add monochrome icon for Android 12+
Android 12+ provides support for monochrome icons in launchers (see below) :

image

It would be a nice touch if Notes also supported this !

salixor:
Uses the existing icon to provide a monochrome variant of the icon, to support Android 12+ themed icons. Closes #1543.

image

On a side note, I've been considering whether or not we should keep the "DEV" text in the monochrome variant. A way around this would be having a single SVG path, rather than several overlaying each other. That way, the DEV text would be preserved. I however haven't found an easy way to do that, and I'm not that skilled in SVGs.

It does still show in places where the themed icon isn't applied :

image


stefan-niedermann:
I see the struggle, but I think it's quite important to keep the `dev` label here for being able to distinguish the versions from each other...

Maybe @jancborchardt is able to do some SVG magic to provide a version of the Notes pencil icon containing the `dev` label?

stefan-niedermann:
@salixor what if we make the `DEV` label just white text and move it below the pen? This should work, no?

salixor:
> @salixor what if we make the `DEV` label just white text and move it below the pen? This should work, no?

Yes, that way it would work ! Here's how it looks :

imageimage

stefan-niedermann:
Awesome, thanks for the contribution!

PR #1546, opened 2022-07-06
Use the new Android 12 SplashScreen API

salixor:
Migration reference : https://developer.android.com/guide/topics/ui/splash-screen/migrate

Here's a before / after for Android 12 with the SplashScreen API (API level 31). Currently, the custom splash screen is ignored due to the new API (and uses the app icon by default) :

| **Before**  | **After** |
| ------------- | ------------- |
| image | image |

And for reference, a before / after for API levels < 31 :

| **Before**  | **After** |
| ------------- | ------------- |
| image | image |


stefan-niedermann:
Awesome! I guess the support lib you added will take care of handling it for older API versions?

salixor:
Yes, it's a backwards compatibility lib Google provides to ease migrating to the new Splash API, and which they use in their migration guide

stefan-niedermann:
- Optional question: Will we need to add the splashscreen also to the `EditNoteActivity`? It can be launched directly, for example from widgets. (I think we are not doing this yet and I am not sure whether this has other implications)
- Will this affect existing installations, for example vanished or unusable launcher links on created at their homescreens?

stefan-niedermann:
> Optional question: Will we need to add the splashscreen also to the EditNoteActivity? It can be launched directly, for example from widgets. (I think we are not doing this yet and I am not sure whether this has other implications)

Currently no changes in those regards has been done. This won't work any better or worse than before the changes, but we might have a look in the future to enhance the splashscreens for such "non-primary" actions

> Will this affect existing installations, for example vanished or unusable launcher links on created at their homescreens?

Tested on an emulated and a physical Android 12 device: Existing shortcuts will continue to work.

Issue #1441, opened 2022-01-04
Card Radius on Widget too big


newhinton:
Please use **[GitHub reactions](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/)** 👍 to show that you are affected by the same issue. Please don't comment if you have no relevant information to add!

**Describe the bug**

The Radius is too big to look good. It comes too close to the content in the corners which looks strange

**To Reproduce**
1. Create Widget with text

**Expected behavior**

A smaller radius would be better

**Screenshots**

![Screenshot_20220104-113544_Trebuchet](https://user-images.githubusercontent.com/25279821/148046790-ebb5bd93-834f-4317-a065-d00f76a3a57c.png)

**Smartphone (please complete the following information):**
 - Nextcloud Notes-Version (android app): 3.4.18
 - F-Droid or Play Store: FDroid
 - Android-Version: Lineageos 18.1
 - Device: 

**Server**
 - Nextcloud version: 
 - Nextcloud Notes version (server app): 

**Stacktrace**

stefan-niedermann:
I agree that it looks like shit, but that's what Android does starting with version 12 (they called it ["improvements"](https://developer.android.com/about/versions/12/features/widgets)).

Since they decided to just crop the widgets to this huge border radius and i won't maintain different widgets per Android version, we'll have to stick with it.

newhinton:
I feared something like that was the reason for the change. Maybe increasing margin_left/margin_right to fit the bottom and top margins will make it look better?

The examples dont look too bad, but they use icons in the corner which helps to hide the issue a bit

newhinton:
On a sidenote, will you support the new compound buttons so that users can change checkbox states from the widget?

jancborchardt:
> Maybe increasing margin_left/margin_right to fit the bottom and top margins will make it look better?

Yeah, that seems like a good improvement.

Also, is it possible to make the top and bottom behave like padding and not like margin – so that the text is only cut off at the widget top and bottom, and not before?

abdusalamApps:
It looks fine on OnePlus Nord with 6.44 inch screen, but looks like the picture provided in the issue on Galaxy S9 with 5.8 inch screen.
However, I created a pull request for this issue. I hope it gets approved and merged
Here: https://github.com/stefan-niedermann/nextcloud-notes/pull/1588

PR #1588, opened 2022-10-08
Issue #1441 - Smaller radius for single note widget

abdusalamApps:
This is a pull request for the change requested in issue 1441 in the original repo


jancborchardt:
@abdusalamApps could you add before/after screenshots for easier design review? Thank you! :)

stefan-niedermann:
I am afraid that this only reduces the border radius ignoring that newer Android versions will hardly crop the border radius furth - leading to unoredictable states...

jancborchardt:
@abdusalamApps could you also make sure to test on newer Android versions as @stefan-niedermann mentioned? Thanks! :)

abdusalamApps:
@jancborchardt I'll do that the first chance I get!

abdusalamApps:
It looks like my change doesn't affect Android 13.0 API 33. See screenshots below:

Android 13.0 API 33, before and after
Android 13 0 API 33 - Before
Android 13 0 API 33 - After

Android 9.0 API 28, before and after
Androdi 9 0 API 28 - Before
Android 9 0 API 28 - After

jancborchardt:
So then considering @stefan-niedermann’s feedback I’m not sure we should merge this. It’s better if we just follow the guidelines as closely as possible. :\

abdusalamApps:
@jancborchardt Could you elaborate on what guide lines to which you're referring?

stefan-niedermann:
> what guide lines

As linked in the issue:

https://developer.android.com/about/versions/12/features/widgets

AndyScherzinger:
Rebased due to major technical updates in terms of libraries, especially AGP and SSO

newhinton:
@stefan-niedermann But the linked docs state that `system_app_widget_background_radius` should be used. I have multiple devices (GrapheneOS, LineageOS) where the notes-widget uses a different radius than every other tested app.

Screenshot_20240401-155424

As you can see the the notes app uses a wider radius. This becomes especially clear when you try to move the widget and the boundaries are shown. (second notes-widget)

The KDE one is also slightly wider than the boundary, but it is imperceptible except for "moving" it where a tiny gap is shown. The wikipedia-one is completely off the rails anyway, but it still uses the correct radius.

I don't know if the proposed solution is the right one, but the current choosen radius is most certainly not the proper one.

newhinton:
I just checked it:
https://github.com/nextcloud/notes-android/blob/cc110fcfe4f72d437c7bcd90f68a1102f68de831/app/src/main/res/values/dimens.xml#L18

This clearly violates the guideline set by google: [the corner radius of the widget background, which is never larger than 28 dp.](https://developer.android.com/develop/ui/views/appwidgets#rounded-corner)

newhinton:
Why not follow the suggestion by the docs?

Use a default value (28dp like you suggested) for skd 30 and below, and

`<item name="backgroundRadius">@android:dimen/system_app_widget_background_radius</item>`

for sdk 31. This should always use the proper radius.

If you like i can create a pr for it.

newhinton:
Wait, sorry. It seems it *is* already using the proper constant. I need to investigate.

stefan-niedermann:
Exactly, the hard coded value is only a fallback for SDKs where `widget_outer_radius` ain't available.

PR #1663, opened 2023-02-01
Add Kotlin support

AlvaroBrey:
As Android team we've agreed to use only Kotlin for new code. Let's get the ball rolling for Notes too.


tobiasKaminsky (approved):

tobiasKaminsky:
Nice! And we have already CI check to prevent new java files.

Issue #1648, opened 2023-01-24
Material 3️⃣ alignment with Files and Talk


AndyScherzinger:
Just to have the discussion, looping in @tobiasKaminsky and @stefan-niedermann what do you think? Should be adapt the styling we put in place for files and talk which I think would basically translate to the top bar(s) and FAB at this point? Can't tell about check-boxes etc. since I didn't take a look at the code / theming yet. First wanted to see if this would even be something to consider before digging deeper and maybe working on a PR. In a way my assumption is it hasn't been put into place to give notes a completely flat UI to mimic paper? (Wild guess, because if Stefan would have wanted it and had the time (don't know) I am sure he would have implemented it already since the UI is slick!)

Looking forward to your comments 💙 

Also @jancborchardt and @nimishavijay for your opinion

stefan-niedermann:
What do you have specifically in mind? The Notes Android app already uses the Material 3 themes 🙂

tobiasKaminsky:
I am all in for alignment, especially code wise, so that we can e.g. also use server/user based theming at some point.
But it has low priority, as app looks already great!

stefan-niedermann:
> use server/user based theming at some point.

I still can't follow you guys, the server theme color is already applied 🙈 - code and optical alignments make sense of course

tobiasKaminsky:
Sorry!
We want to extract theming/color stuff from Talk and Files to android-common and already started with it.
And I think it is best to have then the same code used in Notes.

AndyScherzinger:
Some comparison shots, didn't yet look at the code but running the app against cloud.nextcloud shows the differences to me:

|notes|files|
|---|---|
|![Screenshot_20230126_152209](https://user-images.githubusercontent.com/1315170/214859950-b726f827-d766-4f5c-8be7-9d8511c503ac.png)|![Screenshot_20230125_095346](https://user-images.githubusercontent.com/1315170/214859413-e8912d32-9856-42a4-8716-67955678b053.png)|
|![Screenshot_20230126_152233](https://user-images.githubusercontent.com/1315170/214859968-c8fdf729-1479-4462-9b4d-ee495ac39ff4.png)|![Screenshot_20230125_095433](https://user-images.githubusercontent.com/1315170/214859442-1e987055-80c2-4cdd-8f18-144cb2b41ebb.png)|
[...]

Issue #769, opened 2020-03-22
Search on big notes does not work


chorus12:
Search on big notes does not work

**To Reproduce**
Make a note with 10K+ characters and try making a search of a text in the note.
The interface freezes and you even can not type the word to search. 

**Server**
 - Nextcloud-Version: 16.0.1
 - Nextcloud Notes-Version (server app): 3.2.0
 - PHP-Version: PHP 7.3.6

**Smartphone (please complete the following information):**
 - Nextcloud Notes-Version (android app): 2.5.0
 - F-Droid or Play Store: Play Store 19.2.24
 - Device: Moto Z2 play
 - Android-Version: 8

chorus12:
updated to 2.11 - still freezes and even crashes.

stefan-niedermann:
Well that's no excuse of course and i will try to find a way to make the search more performant, but are you sure you are using the right tool? What kind of notes do you use that take more then 10k characters? Are you writing a book with my app? 😄

chorus12:
Danke. Kind of :) it's a file with daily notes

chorus12:
And as a side note - a control that allows jumping to the end of file or a beginning would be very handy for big files.

stefan-niedermann:
You can just use the upper control at the first occurrence. It will simply wrap back to the last occurrence, then you are at the end.

Isaac-Graham:
Hello there. We are trying to fix this issue. Are there too many repeated search pattern in your note? This issue is reproduced only when the search pattern occurs too many times in the note on my phone/AVD.

stefan-niedermann:
Will be fixed in `v2.11.11`, thanks @Isaac-Graham & team!

Isaac-Graham:
u r welcome : )

PR #1680, opened 2023-02-12
feat(theming): Use nextcloud-common library for theming UI elements (#1648)

stefan-niedermann:
This is an **intermediate** step to an alignment with the Material Files and Talk app.

Most important class: `BrandingUtil` - is more or less what you know as `ViewThemeUtils` in the Files app.

Something more you should know: I created also a class `NotesViewThemeUtils` as a place to put app specific styling stuff until they are available within the `android-common` library. It will probably fail because of the "No more Java" GitHub action you put in place :wink: So either we need to convert it to Kotlin or you make an exception for intermediate steps :man_shrugging: .

- The obvious stuff looks good (`FloatingActionButton`s, `EditText`s, `SwitchPreference`s and so on) and is already done via `android-common`
- Other parts like Toolbars might take some more time to iron out.

If you have own plans or think this goes into the wrong direction, please let me know.

# :eye: Screenshots

Only made with base color `#0082C9` for time reasons, but the other colors look as expected and aligned with the Files app, too.

Light | Dark
--- | ---
grafik | grafik
grafik | grafik
grafik | grafik
grafik | grafik
grafik | grafik

### DIFF
--- a/app/src/main/java/it/niedermann/owncloud/notes/branding/NotesViewThemeUtils.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/branding/NotesViewThemeUtils.java
@@ -1,7 +1,7 @@
 package it.niedermann.owncloud.notes.branding;
 
 import static com.nextcloud.android.common.ui.util.ColorStateListUtilsKt.buildColorStateList;
-import static it.niedermann.owncloud.notes.NotesApplication.isDarkThemeActive;
+import static com.nextcloud.android.common.ui.util.PlatformThemeUtil.isDarkMode;
 
 import android.content.Context;
 import android.graphics.Color;
@@ -21,7 +21,6 @@ import androidx.core.graphics.drawable.DrawableCompat;
 
 import com.google.android.material.appbar.AppBarLayout;
 import com.google.android.material.appbar.MaterialToolbar;
-import com.google.android.material.navigation.NavigationView;
 import com.nextcloud.android.common.ui.theme.MaterialSchemes;
 import com.nextcloud.android.common.ui.theme.ViewThemeUtilsBase;
 import com.nextcloud.android.common.ui.theme.utils.MaterialViewThemeUtils;
@@ -125,7 +124,7 @@ public class NotesViewThemeUtils extends ViewThemeUtilsBase {
 
     @ColorInt
     public int getTextHighlightBackgroundColor(@NonNull Context context, @ColorInt int mainColor, @ColorInt int colorPrimary, @ColorInt int colorAccent) {
-        if (isDarkThemeActive(context)) { // Dark background
+        if (isDarkMode(context)) { // Dark background
             if (ColorUtil.INSTANCE.isColorDark(mainColor)) { // Dark brand color
                 if (NotesColorUtil.contrastRatioIsSufficient(mainColor, colorPrimary)) { // But also dark text
                     return mainColor;
--- a/app/src/main/java/it/niedermann/owncloud/notes/branding/NotesViewThemeUtils.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/branding/NotesViewThemeUtils.java
@@ -42,15 +42,6 @@ public class NotesViewThemeUtils extends ViewThemeUtilsBase {
     }
 
     /**
-     * Use {@link ColorRole#ON_PRIMARY_CONTAINER}
-     */
-    @Deprecated(forRemoval = true)
-    @ColorInt
-    public int getOnPrimaryContainer(@NonNull Context context) {
-        return withScheme(context, Scheme::getOnPrimaryContainer);
-    }
-
-    /**
      * The Notes app uses custom navigation view items because they have several features which are
      * not covered by {@link NavigationItem}.
      */
--- a/app/src/main/java/it/niedermann/owncloud/notes/branding/NotesViewThemeUtils.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/branding/NotesViewThemeUtils.java
@@ -1,5 +1,6 @@
 package it.niedermann.owncloud.notes.branding;
 
+import static com.nextcloud.android.common.ui.util.ColorStateListUtilsKt.buildColorStateList;
 import static it.niedermann.owncloud.notes.NotesApplication.isDarkThemeActive;
 
 import android.content.Context;
@@ -7,6 +8,9 @@ import android.graphics.Color;
 import android.graphics.PorterDuff;
 import android.graphics.drawable.LayerDrawable;
 import android.util.Log;
+import android.view.View;
+import android.widget.ImageView;
+import android.widget.TextView;
 
 import androidx.annotation.ColorInt;
 import androidx.annotation.IdRes;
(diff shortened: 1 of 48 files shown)


stefan-niedermann:
So I will stop here and give you some time to for a review - sorry, it already grew larger than I originally planned...

AndyScherzinger:
The screenshots look good to me 👍

tobiasKaminsky:
Fine by me.
Please check with Alvaro when to merge, as he is also currently working quite a lot on Notes.
Might be that it is wise to postpone this, or merge as early as possible to prevent conflicts.

app/src/main/java/it/niedermann/owncloud/notes/branding/NotesViewThemeUtils.java
```diff
@@ -0,0 +1,157 @@
...
+
+    public NotesViewThemeUtils(@NonNull MaterialSchemes schemes) {
+        super(schemes);
+        this.schemes = schemes;
+    }
+
+    @ColorInt
+    public int getOnPrimaryContainer(@NonNull Context context) {
+        return withScheme(context, Scheme::getOnPrimaryContainer);
+    }
```
AlvaroBrey, line 48: One of the design goals of the ui theming lib is to prevent client apps from having to deal with the schemes directly, instead calling the available theming utils for each specific view type and using `ColorRole`s if they need to choose a specific color. I'll try to figure out what is needed to avoid this method and add it to common:ui.

app/src/main/java/it/niedermann/owncloud/notes/main/MainActivity.java
```diff
@@ -591,21 +594,26 @@ public void onIconClick(NavigationItem item) {
...
+        binding.headerView.setBackgroundColor(color);
+        @ColorInt final int headerTextColor = ColorUtil.INSTANCE.getForegroundColorForBackgroundColor(color);
+        binding.appName.setTextColor(headerTextColor);
+        DrawableCompat.setTint(binding.logo.getDrawable(), headerTextColor);
 
         // TODO We assume, that the background of the spinner is always white
-        activityBinding.swiperefreshlayout.setColorSchemeColors(contrastRatioIsSufficient(Color.WHITE, mainColor) ? mainColor : Color.BLACK);
-        binding.appName.setTextColor(textColor);
-        DrawableCompat.setTint(binding.logo.getDrawable(), textColor);
+        activityBinding.swiperefreshlayout.setColorSchemeColors(contrastRatioIsSufficient(Color.WHITE, color) ? color : Color.BLACK);
```
AlvaroBrey, line 610: There is a `AndroidXViewThemeUtils.themeSwipeRefreshLayout` function available. Was it not working for this specific screen?

app/src/main/java/it/niedermann/owncloud/notes/main/MainActivity.java
```diff
@@ -591,21 +594,26 @@ public void onIconClick(NavigationItem item) {
...
+        final var util = BrandingUtil.of(color, this);
+        util.material.themeFAB(activityBinding.fabCreate);
+        util.platform.colorCircularProgressBar(activityBinding.progressCircular);
+        util.platform.colorNavigationView(binding.navigationView);
+        util.notes.applyBrandToPrimaryToolbar(activityBinding.appBar, activityBinding.searchToolbar, colorAccent);
+
+        binding.headerView.setBackgroundColor(color);
+        @ColorInt final int headerTextColor = ColorUtil.INSTANCE.getForegroundColorForBackgroundColor(color);
+        binding.appName.setTextColor(headerTextColor);
+        DrawableCompat.setTint(binding.logo.getDrawable(), headerTextColor);
```
AlvaroBrey, line 607:
```suggestion
        util.platform.colorViewBackground(binding.headerView, ColorRole.PRIMARY);
        util.platform.colorTextView(binding.appName, ColorRole.ON_PRIMARY);
        util.platform.tintDrawable(this, binding.logo.getDrawable(), ColorRole.ON_PRIMARY);
```

This may accomplish a similar thing to what you wanted in a less roundabout way (needs `ColorRole` import)

AlvaroBrey (changes requested):

app/src/main/java/it/niedermann/owncloud/notes/branding/NotesViewThemeUtils.java
AlvaroBrey: That being said, I'm okay with merging without changing this in the interest of moving forward, but we should add a `@Deprecated` to this method explaining the above.

AlvaroBrey: After https://github.com/nextcloud/android-common/pull/77 is merged, I believe all of the usages of this method can be removed in favor of using `ColorRole.ON_PRIMARY_CONTAINER` where appropriate.

app/src/main/java/it/niedermann/owncloud/notes/main/MainActivity.java
stefan-niedermann: I orientated at the files app, [which does also set the actual account color](https://github.com/nextcloud/android/blob/master/app/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java#L307-L308) and looks like this:

image

With the changes you proposed it would look like this:

image

app/src/main/java/it/niedermann/owncloud/notes/branding/NotesViewThemeUtils.java
stefan-niedermann: I replaced it where possible, but one point is still missing: We are creating `ForegroundColorSpan`s for highlighting the search term in results. I haven't been able to replace this quickly, so I kept the method, added a `@Deprecated` annotation as you suggested.

app/src/main/java/it/niedermann/owncloud/notes/main/MainActivity.java
stefan-niedermann: Just overlooked it :wink:

app/src/main/java/it/niedermann/owncloud/notes/branding/NotesViewThemeUtils.java
AlvaroBrey: Is this enough? https://github.com/nextcloud/android-common/pull/78/files 

If not just leave it as dupe

app/src/main/java/it/niedermann/owncloud/notes/main/MainActivity.java
AlvaroBrey: NP. I have some work to do to improve discoverability of stuff in the ui lib.

AlvaroBrey (approved):
Thank you very much @stefan-niedermann !

stefan-niedermann:
A bit too late, but I just noticed, that this will unresolve #769 - see also https://github.com/nextcloud/android-common/issues/79

It is not a big deal **yet**, because the excerpts are quite small.

I want to use the theming in the `nextcloud-commons` lib too, so the *in* note search aligns with the theme, and *there* it would be quite breaking...

PR #1697, opened 2023-03-01
build: Revert kotlin version to 1.8.0

AlvaroBrey:
1.8.10 not supported by CodeQL for now


tobiasKaminsky (approved):
Reverting is good, and then we merge new codeQL when it is ready, so that we always have a working CI.

PR #1698, opened 2023-03-02
Bump Kotlin to 1.8.10

tobiasKaminsky:
Rebase once codeQL is working with 1.8.10.


AlvaroBrey (approved):
Careful not to merge this; it will pass CI as there is no Kotlin code in `main` for now; but it's not compatible yet

Issue #889, opened 2020-06-20
Start/end padding in lighter black that rest of appbar


thgoebel:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Describe the bug**

In dark mode, the padding at the start/end of the action bar has a slightly lighter color.

**To Reproduce**

1. Open the view/edit screen.

**Expected behavior**

Action bar should be the same color throughout.

**Screenshots**

<img src=https://user-images.githubusercontent.com/33295590/85209623-9efb5280-b339-11ea-8c81-6cba5c24ff08.jpg width=500>

**Smartphone (please complete the following information):**
 - Nextcloud Notes-Version (android app): 2.16.2
 - F-Droid or Play Store: github
 - Android-Version: 10
 - Device: OnePlus 7

**Server**
 - Nextcloud version: 
 - Nextcloud Notes version (server app): 

**Stacktrace**

stefan-niedermann:
Hmm, any idea what's causing it? It's also in `AboutActivity`, `PreferencesActivity` and so on... 

Even if i set `android:background="@color/primary"` on the `AppBar` directly, it has no effect...

thgoebel:
If I had known immediately, I'd have send a PR :)

stefan-niedermann:
Also affects the toolbar of the `NotesListViewActivity` as soon as one clicks on the `MaterialCard`.

stefan-niedermann:
So, i found out that it works if i explicitly set the background color of the `AppBarLayout` programmatically.
This is good enough as a workaround for me, [i left a `FIXME` in the code](https://github.com/stefan-niedermann/nextcloud-notes/commit/07735922bada3583d702c3abf3bcf2c5f4e592ab#diff-20a39f1933af0977be2ea042436cf376R56) so one can have a closer look whenever one has time for racking brains why the hell android does not accept the declarative background color.... :confused: 

Thanks for reporting :slightly_smiling_face:

PR #1752, opened 2023-04-07
feat(theming): Align theming closer to files app

stefan-niedermann:
Follow-Up to https://github.com/nextcloud/notes-android/pull/1680

- Background color is now `@color/bg_default`, like in the Files app
- Statusbars and Toolbars are now themed, except on the main view
- I tackled some technical debts (odd horizontal paddings in toolbars, see #889)

Open issues:

- `SettingsActivity`: Back arrow is not perfectly aligned with icons. I propose to just keep it as it is for now.
- The toolbar icons do have a different position in the main activity compared to each other activity - because of placing it *inside* of the rounded card. This behavior will be the same when we switch to the material `SearchBar`
- ActionMode could need some more love

Out of scope:

- Switching to an actual material `SearchBar` for the main view
image

Would love to get some feedback @AndyScherzinger


AndyScherzinger (approved):
looks good to me @stefan-niedermann 👍

AndyScherzinger:
Minor detail, the 3-dot overflow menu icons should be tinted the same way the other trailing icons are tinted. 

see:

https://github.com/nextcloud/android-common/blob/08db28c1a4ca825c82ddf56747d7892c719550a7/ui/src/main/java/com/nextcloud/android/common/ui/theme/utils/MaterialViewThemeUtils.kt#L58-L62

AndyScherzinger:
> Minor detail, the 3-dot overflow menu icons should be tinted the same way the other trailing icons are tinted.

Well, it is also like that in the talk app, so maybe keep it as is for the moment 👍

stefan-niedermann:
> the 3-dot overflow menu icons should be tinted the same way the other trailing icons are tinted.

Maybe `themeToolbar`  should call `colorToolbarIcon` implicitly? It would fix also the Talk app then and would at least match my persobal expectations...

AndyScherzinger:
i now fixed in for talk explicitly but yeah, makes sense to call it within the toolbar theming logic 👍

AndyScherzinger:
android-common 0.8.0 ships a fix for theming the trailing icons (overflow icon) when theming the toolbar. 👍

stefan-niedermann:
:+1: Ready to merge from my side

PR #1771, opened 2023-04-17
Streamline commons version

tobiasKaminsky:
closes #1769 
closes #1768 
closes #1767 
closes #1763


stefan-niedermann (commented):
LGTM, can be merged.


Hint: Missing the `QS` version of the files app. The behavior is identical to before this PR, just a note, that in the future it might make sense to iterate the enum entries (or better check which account type is actually used)

AndyScherzinger (approved):

PR #1773, opened 2023-04-17
Allow to remove all checked items (fix #1521)

### LINKED ISSUE #1521: Add a way to delete all checked items
Hi,

first of all, many thanks for the app!

**Is your feature request related to a problem? Please describe.**

This is a feature request.
I mainly use the app to maintain lists (groceries, todo lists, etc.).
I'd like a way to quickly delete all the checked items.

**Describe the solution you'd like**

In the 3 dots menu, add an action "Delete all checked items from lists". When clicked, every item in a list, that is checked, is deleted.

nicofrand:
Hi,

This PR solves the issue described in issue #1521 (ie. I want a quick action to clear the markdown from all checked items). It's quite simple but I am new to this language and environment so please tell if there are things that feel off.

Thanks in advance for the review, and thanks again for this useful app!


app/src/main/java/it/niedermann/owncloud/notes/shared/util/NoteUtil.java
```diff
@@ -123,6 +126,29 @@ public static String getLineWithoutMarkdown(@NonNull String content, int lineNum
...
+        );
+        final Matcher matcher = pattern.matcher(content);
+
+        // We can't just replace the matcher with an empty string as this would
+        // let a blank line. To be able to remove the blank line as well, we set
+        // a random-ish token that we will then match with the EOL as well and
+        // remove them all together.
+        return matcher
+                .replaceAll("RANDOM-ISH_TEXT_TO_REPLACE")
+                .replaceAll("RANDOM-ISH_TEXT_TO_REPLACE\n?", "");
```
stefan-niedermann, line 149: I don't have a better idea, but this looks.... odd...

app/src/main/java/it/niedermann/owncloud/notes/shared/util/NoteUtil.java
```diff
@@ -123,6 +126,29 @@ public static String getLineWithoutMarkdown(@NonNull String content, int lineNum
...
+     *
+     * @param content    String
+     * @return newContent String
+     */
+    @NonNull
+    public static String getContentWithoutCheckedItems(@NonNull String content) {
+        final Pattern pattern = Pattern.compile(
+                "^ *[-+*] \\[[xX]\\].*$",
+                Pattern.MULTILINE
+        );
```
stefan-niedermann, line 140: This omits some variants that could lead to undefined behavior, for example checkboxes within code blocks. You can get some inspiration how to handle codeblocks in the [`nextcloud-commons`](https://github.com/stefan-niedermann/nextcloud-commons/blob/master/markdown/src/main/java/it/niedermann/android/markdown/MarkdownUtil.java#L140) library, but beware: It ain't that easy. Codeblocks can have three or more backticks, for example

app/src/test/java/it/niedermann/owncloud/notes/shared/util/NoteUtilTest.java
```diff
@@ -99,4 +99,40 @@ public void testGenerateNoteExcerpt_sdk_30() {
...
+
+        String expected = "- [ ] Line A"
+                .concat(newLine)
+                .concat(newLine)
+                .concat("* [ ] Line E")
+                .concat(newLine)
+                .concat("+ [ ] Line G")
+                .concat(newLine)
+                .concat("    + [ ] G1")
+                .concat(newLine);
```
stefan-niedermann, line 134: As stated above, there are very many cases of the `commonmark` specification not considered (and also not covered by the tests)... @AndyScherzinger  and @tobiasKaminsky will have to decide how compliant is "compliant enough" :wink:

app/src/main/java/it/niedermann/owncloud/notes/shared/util/NoteUtil.java
nicofrand: Thanks, I will!

app/src/main/java/it/niedermann/owncloud/notes/shared/util/NoteUtil.java
nicofrand: I know but I could not find a better way either

stefan-niedermann:
@AndyScherzinger you also maybe want to talk with @jancborchardt or @nimishavijay about the UX? The overflow menu is already quite crowded

nimishavijay:
Thanks for the contribution @nicofrand :) Could you please post a screenshot or video of the UI so that it is easier for everyone to review the design? :)

nicofrand:
Sure, here it is:

[device-2023-05-19-154243.webm](https://github.com/nextcloud/notes-android/assets/3419050/0c495fe7-2bf5-48f1-9694-50273e51c46f)

jancborchardt:
Very nice contribution @nicofrand! :) I’m wondering if instead of a permanent action like this, it would be good to have one of either of these:
- Automatically sort the checked items towards the bottom (of course within their hierarchy) → iOS Notes does this and it’s super useful e.g. for shopping lists and things like that
- The menu item could be a checkbox instead of a button, saying "Hide checked items"

What do you think @stefan-niedermann @nextcloud/designers?

stefan-niedermann:
We had this feature request a few times (hide or strike checked items, sort checked items to bottom), which I declined so far because
- feature parity with the Notes server app¹
- The scroll state of the preview and edit mode (which is restored when switching modes) would be different depending on the amount of checked items²
- for sorting:
  - we didn't want to manipulate the content automatically in the background
  - we have no clear idea how to handle multiple lists separated by other elements

¹ Might be obsolete now when you implement it on server as well
² Maybe this should be considered for the rich editor as well

After all it's your decision now, and maybe some advanced handling of checkbox lists is indeed a good idea, but you should think about consequences, especially parity between the modes (view, edit, rich), parity with the ecosystem (notes server, notes iOS) and parity with other occurrences of the rich editor (files, deck, tables, polls, collectives, ...)

nicofrand:
Hi everyone!

So, if I sum up the concerns regarding this merge request, here are the points I see:

1. This would also erase checked items inside code blocks. I think we could solve this by checking if the number of backticks from the beginning of the document until the beginning of the checkbox is impair (if there is 1 or 3 back ticks, we are in a code block, else we are not)?
2. Design: maybe this should appear in another menu. It seemed appropriate for me to put this here, specially since there are no other menus but this can be discussed of course
3. This should first be implemented in nextcloud notes itself. I am afraid I wouldn't have the time to do both :/. It would be great if an issue with all concerns solved first existed and just needed an implementation I guess.
4. Replace the deletion with a hidden state. I don't see how that would help me as a user with a shopping list for ex. Moving items to the end would not reduce the size of the list. And what about duplicates then? Should the item be moved to the end of the document, end of the list, end of the sublist? I feel like this is a different feature…

Could you guys tell me if I should move forward to solve point 1 or if there is no current change of merging this?

jancborchardt:
I would say if it’s the simple hiding option like so:

> The menu item could be a checkbox instead of a button, saying "Hide checked items"

It would be good, and it would also not necessarily also be implemented in the web app too.

@nicofrand what do you mean by it wouldn't help you with a shopping list? Usually shopping lists are similar across shopping runs, otherwise you have to type all deleted items again and again.

@stefan-niedermann what do you think?

jancborchardt:
> Having it in a bottom list, usually with a stroke, often creates duplicates (I don't remember if I already added it, where in the list, as a plural or not, etc.).

With the hiding option I mean that the items should actually not be moved around in the list, just hide checked items. :) So a purely client-side view switch. Does that make sense?

AndyScherzinger:
Rebased due to major technical updates in terms of libraries, especially AGP and SSO

Issue #927, opened 2020-07-28
Content visble behind biometric 


S-360:
I have setup a biometric lock. Why is the content of the app visible behind the lock? Shouldn't the content of the note app be masked until the app is successfully unlocked?

stefan-niedermann:
It is supposed to be hidden, yes. This worked until Android Q, until Google decided to change the behavior. I don't know yet how to restore the old behavior, if you have an idea: Pull Requests are welcome 😉 

cc @beatbrot and @dan0xii

S-360:
That's strange? all other appson my device hide content until theapp is unlocked via biometrics.

Running Android 10 on Samsung A50

stefan-niedermann:
Maybe we have to launch another (blank) activity which itself prompts the biometric screen 🤔 This way the content would stay hidden, i guess. Will have to try this, result handling will ne a pain in the ass :-/

carabao72:
> I have setup a biometric lock. Why is the content of the app visible behind the lock? Shouldn't the content of the note app be masked until the app is successfully unlocked?

Personally, to avoid displaying the note before fingerprint unlocking, I use a little trick by putting a line of 'xxxxxxxxxxxxxxxx' just after the title of the note.  So I see before unlocking: note title followed by xxxxxxxxxxxx

vhouriet:
It's probably the same issue or related to that. But if you open the app and don't unlock it and then open the overview, you can see the notes as well in the overview.

![Screenshot_20211005-141518_Trebuchet](https://user-images.githubusercontent.com/89147152/136020407-d2d81d30-b306-488f-bc2e-af76766ac421.png)

I feel like local encryption should fix the problem. Even if the content is visible, it will only be gibberish then.

stefan-niedermann:
@vhouriet no, this has been fixed in `v3.3.0` (not yet available at the Play Store) with a setting to disable screenshots (See https://github.com/stefan-niedermann/nextcloud-notes/issues/1098). This is unrelated to the biometric issue.

PR #1905, opened 2023-09-26
Hide notes when app is locked

F462:
In this PR, the notes are not loaded before the app is unlocked, which one the one hand still preserves the app impression while locked, while at the same time hiding the sensitive data. This solves #927 .

Since I'm quite new to the code, I cannot really judge if this approach is valid. However, it seems to achieve the expected result without side effects (tested on Android 13).

The notes are still shown in the app preview when minimized, which is a separate issue.

In addition, the time based app locking doesn't seem to work on the main screen, which isn't supported by these changes here, either.


stefan-niedermann:
Hmm, might work (didn't check it due to missing sparetime at runtime).

However, looking at three different banking apps it seems an dedicated `UnlockActicity` seems the way to go to solve Googles shit show eventually. 🤷

cc  @tobiasKaminsky and @AndyScherzinger for their opinion

AndyScherzinger:
I agree @stefan-niedermann with the assessment regarding a dedicated activity. This is also how we do it on files and talk.

PR #2011, opened 2024-02-07
Use ConnectivityManager Instead Deprecated IntentFilter(ConnectivityManager.CONNECTIVITY_ACTION) and BroadcastReceiver

alperozturk96:


alperozturk96:
@stefan-niedermann  I use ConnectionLiveData two times due to updateNetworkStatus() usage. I think we can remove it.

**updateNetworkStatus()** this function gets called from MainViewModel. Can we remove it and only use syncObserver?

JonasMayerDev (approved):
In quick test, looks good to me

PR #2103, opened 2024-04-01
Test

AndyScherzinger:
Just to show to @stefan-niedermann


AndyScherzinger:
So latest master works for me @stefan-niedermann using
https://jitpack.io/com/github/stefan-niedermann/nextcloud-commons/1ed158ba44/build.log

Hence the update dependency PR since the updating to DynamicScheme broke the lib for me for markdown due to the manual update of the common lib to 0.19.0 in the build file.

PR #2114, opened 2024-04-04
feat(meta): Migrate issue templates to issue forms

stefan-niedermann:
I observed an increasing number of issues with extremely poor quality.  
I expect especially less cases where the issue template simply gets ignored or dropped.


AndyScherzinger (approved):
LGTM 👍

PR #2115, opened 2024-04-04
Use FLAG_IMMUTABLE if needed

tobiasKaminsky:


tobiasKaminsky:
Tested with Android 10 and 14.

PR #2209, opened 2024-05-23
PoC: Provides toolbar for native markdown editor

stefan-niedermann:
Currently the toolbar is positioned below the content so
- the actions are reachable more ergonomically and
- the additional vertical gap when switching between preview and edit mode can be avoided
However, the collaborative rich editor shows its menu bar above the content.

Which way should we go? Consistency vs. ergonomics? Or would you even consider to switch the toolbar position for the collaborative rich editor?

CC @tobiasKaminsky @AndyScherzinger

| _ | _ |
| --- | --- |
| Screenshot_20240523_113405 | Screenshot_20240523_114309 |

Please note that the toolbar integration *might* still have some bugs left which would need to be addressed in the `stefan-niedermann:nextcloud-commons` library.

### DIFF
--- a/app/src/main/java/it/niedermann/owncloud/notes/branding/NotesViewThemeUtils.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/branding/NotesViewThemeUtils.java
@@ -13,6 +13,7 @@ import android.content.Context;
 import android.content.res.ColorStateList;
 import android.graphics.Color;
 import android.graphics.PorterDuff;
+import android.graphics.drawable.Drawable;
 import android.graphics.drawable.LayerDrawable;
 import android.util.Log;
 import android.view.View;
@@ -26,6 +27,8 @@ import androidx.appcompat.widget.AppCompatAutoCompleteTextView;
 import androidx.appcompat.widget.SearchView;
 import androidx.appcompat.widget.Toolbar;
 import androidx.core.content.ContextCompat;
+import androidx.core.graphics.BlendModeColorFilterCompat;
+import androidx.core.graphics.BlendModeCompat;
 import androidx.core.graphics.drawable.DrawableCompat;
 
 import com.google.android.material.appbar.AppBarLayout;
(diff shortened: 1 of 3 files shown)


AndyScherzinger:
Hmmm, good question.

I checked the notes apps of Google and OnePlus and both have the editor bar at the bottom.

@nextcloud/designers what do you think?

marcoambrosini:
Great addition :) 
Out of these two I would prefer the bar at the bottom. 

However, I would favor a floating toolbar that is not shown at all times, but only when text is selected and an editing operation is likely to occur. 

As an example, this is the toolbar that pops up in signal when text is selected.

Screenshot_20240527-091645

stefan-niedermann:
> However, I would favor a floating toolbar that is not shown at all times, but only when text is selected and an editing operation is likely to occur.

This describes exactly the same behavior as today (see [FAQ](https://github.com/nextcloud/notes-android/blob/main/FAQ.md#why-arent-there-any-buttons-to-apply-formatting)) and it will continue to work. However there are some drawbacks:

1. This behavior has never been fully accepted by the users. A quick search in the issue section will show you dozens of requests to always have buttons present
2. This behavior is not supported by Nextcloud Text and therefore would contradict the efforts to minimize the inconsistences between the two edit modes of the Notes Android app as well as the inconsistences of Notes Android compared to Notes server

marcoambrosini:
1. I think it might be a vocal minority of users. For the purpose of taking notes on mobile I think it's perfectly fine if text style is not always visible.
2. I don't think we should aim at "consistency at all costs" on different platforms. I would favour displaying more of the content of the note on small screens over the pursuit of consistency in this case. Also, on this point I'd very much like to see this "edit on select" behaviour coming on the text app on the web. It's just overall more intuitive and requires less jumping around with the mouse.

That said, if you really think we need this, I vote for edit bar just above the keyboard :)

stefan-niedermann:
> I don't think we should aim at "consistency at all costs" on different platforms

It's about consistency within the app itself. The Notes Android app ships two editors which can be toggled in the Notes Android apps settings, both with pros and cons.

> Also, on this point I'd very much like to see this "edit on select" behaviour coming on the text app on the web

This has been declined in https://github.com/nextcloud/notes/issues/200#issuecomment-1480145991

Additional context: A [similar issue](https://github.com/nextcloud/notes/issues/788) exists for the Notes server app (which also ships two editors).

larena1:
@jancborchardt have you instructed your developers to abandon this app or what is going on? Latest release 1 year ago?

jancborchardt:
@larena1 thanks for caring and asking – as you might have seen, there have been steady releases since your comment, and also a recent release candidate at https://github.com/nextcloud/notes-android/releases
Please help testing that if you can, would be much appreciated! :)

larena1:
@jancborchardt yes the situation seems to have improved a bit after one year without releases

Unfortunately , the changelog is overwhelmingly filled with automated Renovate commits for minor dependency updates, making it difficult to identify meaningful changes and test them effectively. Maybe these could be highlighted better?

jancborchardt:
@larena1 good point – @alperozturk96 maybe it would be nice to filter out all these automated commit messages, to only focus on bug fixes and features. :)

AndyScherzinger:
The changelog is auto-generated by Github but I took the time to better structure it, only did it for https://github.com/nextcloud/notes-android/releases/tag/rc-4.5.0-01 though

Issue #1729, opened 2023-03-26
Slow search in long notes


MiMoHo:
Please use **[GitHub reactions](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/)** 👍 to show that you are affected by the same issue. Please don't comment if you have no relevant information to add!

**Describe the bug**
When searching for a term in a very long note, the app doesn't wait for the user to finish typing but starts searching instantly which results in finding matches for the first letter and then recognizing that the user typed a much longer term and starts searching again. 

**To Reproduce**
1. Create a new note, add a novel of text to it and save by going back to the overview.
2. Reenter the note and type a word in the search bar.
3. First, the app will only show one letter in the search bar, but you can keep typing on the keyboard. Do so.
4. The first search iteration will highlight all findings for the first letter of the word and then starts over instantly again for the full word.

**Expected behavior**
There should be a search delay of one second after the user stopped typing before the search is triggered. This will result in cutting down the search time by half in long notes.

**Screenshots**
[Search results after the first iteration while the second iteration is running and will produce findings for the entire word soon](https://user-images.githubusercontent.com/37556964/227781000-1c7a79bd-3e3c-46fa-ac37-79df6959a7f3.png)

**Smartphone (please complete the following information):**
 - Nextcloud Notes-Version (android app): 4.0.0 RC1
 - F-Droid or Play Store: FDroid
 - Android-Version: 11
 - Device: Sony Xperia

**Server**
 - Nextcloud version: 26.0.0
 - Nextcloud Notes version (server app): 4.7.2

**Stacktrace**

stefan-niedermann:
> There should be a search delay of one second after the user stopped typing before the search is triggered. This will result in cutting down the search time by half in long notes.

No, there shouldn't. There already *is* a [delay of `50ms`](https://github.com/nextcloud/notes-android/blob/main/app/src/main/java/it/niedermann/owncloud/notes/edit/SearchableBaseNoteFragment.java#L42). Waiting for a whole second before starting a simple search in any note for each use case would be terrible.

> long notes

Could you please define your long "note"? Are we talking about 100 characters? 1k? 10k?  100k? Maybe you should consider to split your book up into multiple smaller notes?

[...]

PR #2348, opened 2024-08-28
Fix debouncing not working on first search character; add extra delay instead

cooperra:
For some unknown reason, debouncing was specifically avoided for the first character. This is weird because it's where debouncing is needed most, especially for long files where a single-character search can hang the app for a long time.

I don't think this will fix everything about long-running searches, but debouncing all characters will at least match expectations, allowing people to attempt to type quickly to avoid them.

Since 50ms seemed very quick (a three-frame window at 60 fps), I decided to use a longer delay (200ms) for short searches three characters long or less. The original delay is retained for other lengths, where interactive incremental behavior is more useful and won't hang the app if it happens accidentally.

Should partially address #1729 and #769.


ZetaTom (approved):
Thank you for this contribution to our project!

During my testing, everything worked as expected. This change does in fact improve the user experience of the search function.

ZetaTom:
@cooperra, I've just noticed that one of your commits isn't signed off properly. Please follow the [instructions provided by DCO](https://github.com/nextcloud/notes-android/pull/2348/checks?check_run_id=29339832395), so that we can merge this pull request.

PR #2487, opened 2024-11-14
Add sort lexicographical descendant for notes application

Kokika:


tobiasKaminsky:
Thank you very much for your contribution! :tada: 

I tested it a bit and found that the "sorting header" is off:


Note "Sync" should be in "S", while the "New note…" should be in "N".
This works correct with ascending sort order.

Kokika:
Thank you for testing the PR. I fixed the SectionI Item position at ordering. Could you check the fix?

tobiasKaminsky (approved):

PR #2603, opened 2025-03-18
fix pendingIntent

tobiasKaminsky:
- open note
- pin to homescreen


alperozturk96 (changes requested):
I suggest to update explanation of the function.

[update_explanation.patch](https://github.com/user-attachments/files/19315679/update_explanation.patch)

I can't suggest changes directly from GitHub because I can only able to select changed lines.

PR #2684, opened 2025-06-25
Update Libs

alperozturk96:


tobiasKaminsky (changes requested):
Tobi: change APG version, otherwise AndroidStudio fails

PR #2704, opened 2025-07-01
Migrate to outline icon style

AndyScherzinger:


alperozturk96:
Looks good to me, but I think share internal link icon needs to be changed?


<table>
  <tr>
    <td>Image 1</td>
    <td>Image 2</td>
    <td>Image 3</td>
  </tr>
</table>

AndyScherzinger:
> Looks good to me, but I think share internal link icon needs to be changed?

@jancborchardt what do you think? https://github.com/nextcloud/notes-android/pull/2704#issuecomment-3026760596 it is the icon we also use in all other clients and web at the moment, not sure there is an alternative we want to switch to. It is an outlined icon already after all.

jancborchardt (commented):
> > Looks good to me, but I think share internal link icon needs to be changed?
> 
> @jancborchardt what do you think? [#2704 (comment)](https://github.com/nextcloud/notes-android/pull/2704#issuecomment-3026760596) it is the icon we also use in all other clients and web at the moment, not sure there is an alternative we want to switch to. It is an outlined icon already after all.

Yeah, the icon looks good, I’m a bit confused as to why it would have to be changed? It’s outlined as you say.

alperozturk96:
> > > Looks good to me, but I think share internal link icon needs to be changed?
> > 
> > 
> > @jancborchardt what do you think? [#2704 (comment)](https://github.com/nextcloud/notes-android/pull/2704#issuecomment-3026760596) it is the icon we also use in all other clients and web at the moment, not sure there is an alternative we want to switch to. It is an outlined icon already after all.
> 
> Yeah, the icon looks good, I’m a bit confused as to why it would have to be changed? It’s outlined as you say.

Screenshot 2025-07-02 at 14 15 21

Yes, the icon is outlined, but the circle background for the internal share link doesn't match the rest of the screen. I think having an outlined icon with filled background looks weird.

I just noticed that, but we can skip it (or  for now), since all the icons have been replaced correctly and the app is running without issues.

Thanks for the input.

PR #2715, opened 2025-07-08
Add Debug APK Creation

alperozturk96:


AndyScherzinger (approved):

alperozturk96:
Workflow artifacts are not publicly accessible, I have to check it.

AndyScherzinger:
> Workflow artifacts are not publicly accessible, I have to check it.

@alperozturk96 but org members? If so at least GH org folks especially designers would have access

jancborchardt (commented):
Nice! Just as possible improvements, since now it’s a bit more cumbersome than before:
- [ ] It would be nice to get a direct link to the file, instead of going to the Artifacts section on the Actions page
- [ ] The file is a zip, would be better as apk so you don’t need to fiddle with it
- [ ] It would be nice if the debug build doesn’t overwrite one’s productive app but has a different identifier
- [ ] The link https://github.com/nextcloud/notes-android/actions/runs/16463464562 is publicly accessible, but there’s no way to download (e.g. in private browser tab). Not sure if it needs just a GitHub account or being in the Nextcloud org

alperozturk96:
> Nice! Just as possible improvements, since now it’s a bit more cumbersome than before:
> 
> * [ ]  It would be nice to get a direct link to the file, instead of going to the Artifacts section on the Actions page
> * [ ]  The file is a zip, would be better as apk so you don’t need to fiddle with it
> * [ ]  It would be nice if the debug build doesn’t overwrite one’s productive app but has a different identifier
> * [ ]  The link https://github.com/nextcloud/notes-android/actions/runs/16463464562 is publicly accessible, but there’s no way to download (e.g. in private browser tab). Not sure if it needs just a GitHub account or being in the Nextcloud org

 I'm already aware of those details and planning to work on it. Thanks for the feedback.

tobiasKaminsky:
The zip part is the reason, why in Files repo I put this on my private server.
(we can also do it here)

tobiasKaminsky:
New plan:
- move files QA workflow to android-config
- this way it can be used with Talk, Notes and Files

PR #2901, opened 2025-11-03
fix: status emptiness check

alperozturk96:
<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>before</td>
    <td>after</td>
  </tr>
</table>


AndyScherzinger (approved):
works like a charm, I took the liberty to add a cut-out effect on the status, like we have on web and other clients

PR #2944, opened 2025-12-03
Rezise single line list items

AndyScherzinger:
follow-up for https://github.com/nextcloud/notes-android/pull/2940#issuecomment-3602267799

|Old|Current|New|
|---|---|---|
|old_list_no_selection|before|Screenshot_20251209_124621|

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,51 @@
+# .editorconfig
+
+# see http://EditorConfig.org
+
+# SPDX-FileCopyrightText: 2018-2025 Nextcloud GmbH and Nextcloud contributors
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+# This is the file in the root of the project.
+# For sub folders you can have other files that override only some settings.
+# For these, this settings should be false.
+root=true
+
+[*]
+max_line_length=120
+# use spaces, not tabs.
+indent_style=space
+indent_size=4
+
+[*.yml]
+max_line_length=150
+
+charset=utf-8
+
+# Trimming is good for consistency
+trim_trailing_whitespace=true
+# I've seen cases where a missing new_line was ignored on *nix systems.
+# Never again with this setting!
+insert_final_newline=true
+
+[*.properties]
+# Exception for Java properties files should be encoded latin1 (aka iso8859-1)
+charset=latin1
+
+[*.{cmd,bat}]
+# batch files on Windows should stay with CRLF
+end_of_line=crlf
+
+[*.md]
+trim_trailing_whitespace=false
+
+[.drone.yml]
+indent_size=2
+
+[*.{kt,kts}]
+ktlint_code_style = android_studio
+# IDE does not follow this Ktlint rule strictly, but the default ordering is pretty good anyway, so let's ditch it
+ktlint_standard_import-ordering = disabled
+ktlint_standard_no-consecutive-comments = disabled
+ktlint_function_naming_ignore_when_annotated_with = Composable
+ij_kotlin_allow_trailing_comma = false
+ij_kotlin_allow_trailing_comma_on_call_site = false


kra-mo:
> Title-only notes are pretty rare (I’d say), so they shouldn’t look out of place.

Good point, makes sense.

Just make sure that the right/bottom padding is the exact same as tags look odd otherwise.

And if the size is unified anyway, there is no need to move tags to the left of the date for title-only notes, just keep them below the date as well. Maybe even keep the title top-aligned to then match with the date and better communicate that the note is empty.

TL;DR I’d keep the exact same layout for title-only notes.

jancborchardt (changes requested):
Agree with @kra-mo on the whitespace comment (compact is a bit too tight).

Regarding the emptiness of empty notes, could we vertically center the title of those? That’s commonly done with those classic 2-line elements when there’s no subline.

AndyScherzinger:
@kra-mo @jancborchardt I changed there padding to `12dp`so between the very dense `8dp` and the less compact `16dp`. See below the 3 variants I did for dash and "no content" also in italic (could also discuss to use a different font color (some grey-ish tone) for no content.

|12dp|12dp dash|12dp no content|12dp no content _italic_|12dp unified|
|---|---|---|---|---|
|12dp|12dp_dash|12dp_noContent|12dp_noContent_italic|12dp_unified|

AndyScherzinger:
@jancborchardt also added a 12dp unified version

kra-mo:
> could we vertically center the title of those?

I mean that's what was done previously, but it made it less easy to parse to my eyes.

@AndyScherzinger iirc we avoid italics for a11y reasons. I think the non-italic "No content" one works well enough here :)

kra-mo:
> @jancborchardt also added a 12dp unified version

Yeah, that shifts the layout around too much I feel.

jancborchardt:
> I mean that's what was done previously, but it made it less easy to parse to my eyes.

@kra-mo in practice, I have 0 notes of my 130 which are empty. I’d be curious if having a bunch of empty notes is even something that happens often.

But even if: As said we do the same in our normal components that we use e.g. in Contacts, and it looks much cleaner. Also it gives an implicit indicator about "this is empty" rather than an explicit "No content" text.

@AndyScherzinger one point on the 12dp unified: Is it possible to only vertically align the title, but keep the date and tag 2 lines on the right?

AndyScherzinger:
I am fine with any, just need to know which so we can have it merged. The unified still would need some adjustments, to have things properly aligned (headline start position) and generally the section headers.

So whatever variant gets chosen I'll fix these aspects upfront before merging of course 👍

kra-mo:
> @kra-mo in practice, I have 0 notes of my 130 which are empty. I’d be curious if having a bunch of empty notes is even something that happens often.

I mean sure? But that just sounds more like "this question does not matter". At least I don't see how this validates one approach over another.

> But even if: As said we do the same in our normal components that we use e.g. in Contacts, and it looks much cleaner. Also it gives an implicit indicator about "this is empty" rather than an explicit "No content" text.

My main problem with centering is it very much makes it look like a different *type* of item, which it is not. We shouldn't forget that this is a smaller representation — a preview — of an existing, actual note, whose layout the centering does not mirror: It goes from top to bottom.

In the case of Contacts for example, it is a *different* representation, one with less info, rather than an extract which ellipsizes.

I think Mail would be a much better comparison here as instead of fields and different data points, they both have arbitrarily long text which needs to be trimmed. There, when there is no subject, there is a placeholder "No subject" label as well.

AndyScherzinger:
Here is the version @jancborchardt suggests with the right-side meta data being aligned the way it is done on a 2-line item

12dp_unified_rightMetadataAligned

jancborchardt:
> I think Mail would be a much better comparison here as instead of fields and different data points, they both have arbitrarily long text which needs to be trimmed. There, when there is no subject, there is a placeholder "No subject" label as well.

Good point @kra-mo – let’s go with 12dp "No content" then. :)

AndyScherzinger:
@kra-mo @jancborchardt please re-review, I updated the screenshots in the issue description to reflect the code change, as discussed, see https://github.com/nextcloud/notes-android/pull/2944#issue-3689064475

@alperozturk96 for technical review (I also did some cleanup for the dimensions and added an editorconfig)

PR #2962, opened 2025-12-18
Make external/app sharing within Android more clear and platform-aligned

AndyScherzinger:
*Language mix is due to the lack of translations given it is a new string.*

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
before|after


jancborchardt (approved):
As discussed I’d prefer if it’s at the end of the list (and probably our "share link" should be a button as well?) but that’s more difficult and can be tackled separately.

This is a nice improvement. :)

AndyScherzinger:
As discussed in the chat let's leave it as is for the moment, the other items are part of the list itself, so it is technically doable to also add this to the recycler view I think but additional work larger than just the simple visual change of the PR.

Let's tackle it separately 👍

PR #2974, opened 2025-12-23
Sharing fixes

tobiasKaminsky:


app/src/main/java/it/niedermann/owncloud/notes/persistence/sync/ShareAPI.kt
```diff
@@ -67,6 +67,6 @@ interface ShareAPI {
     @GET("shares/?format=json")
     fun getShareFromNote(
         @Query("path") path: String,
-        @Query("shared_with_me") sharedWithMe: Boolean = true
+        @Query("shared_with_me") sharedWithMe: Boolean = false
```
AndyScherzinger: The web calls

```
format=json
path=%2FNotes%2FSamples%2FFormatting.md
shared_with_me=true
```
So I am not sure if this is correct to be changed

alperozturk96:
@tobiasKaminsky 

What issue or bug does this PR fix? What should we test, and how should we test it? Could you share steps to reproduce the issue?

app/src/main/java/it/niedermann/owncloud/notes/persistence/sync/ShareAPI.kt
alperozturk96: By default is `true` in current master but expected value is `false` then? And in where exactly this didn't set to actual value.

app/src/main/java/it/niedermann/owncloud/notes/share/adapter/holder/ShareViewHolder.java
```diff
@@ -78,7 +77,7 @@ public void bind(OCShare share, ShareeListAdapterListener listener) {
                 binding.icon.setTag(share.getShareWith());
 
                 if (share.getSharedWithDisplayName() != null) {
-                    AvatarLoader.INSTANCE.load(context, binding.icon, account, share.getSharedWithDisplayName());
+                    AvatarLoader.INSTANCE.load(context, binding.icon, account, share.getShareWith());
```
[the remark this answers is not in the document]
alperozturk96: Following string will be used to fetch avatar. username is expected value.

`url + "/index.php/avatar/" + Uri.encode(username) + "/64"` 

When `getSharedWithDisplayName()`, `getShareWith()` these two returning different values?

Image

PR #3043, opened 2026-02-10
fix(share): permission

alperozturk96:
### Fixes

Updates the view to correctly reflect share fields.
De-duplicates share fetch logic and standardizes its usage.
Fixes share update logic.
Fixes sending share emails.
Fixes displaying email addresses in the search bar.
Fixes icon alignment issues.
Replaces buttons with Material 3 icon buttons.
Removes non-existent actions for notes.
Warns users when selecting a weak share password instead of failing silently.
Ensures cached share data is removed when a share is deleted.
Clarifies function naming to distinguish between `get` and `fetch`.
Fixes update POST request body and aligns it with the web implementation.
Unifies share permission setting and checking logic.
Implements allow download and sync option for `Internal` and `User` share type just like web

### Public share fields tests

https://github.com/user-attachments/assets/fbe18a69-9558-4229-b3be-f67517f9386d

### Public share set password

https://github.com/user-attachments/assets/8de5f177-9c8a-4498-8527-90c7cf89b13e

### Internal share tests

https://github.com/user-attachments/assets/4fa7884b-a526-4b62-af3e-fcbacd183e7b

### DIFF
--- a/app/src/main/java/it/niedermann/owncloud/notes/share/NoteShareDetailActivity.kt
+++ b/app/src/main/java/it/niedermann/owncloud/notes/share/NoteShareDetailActivity.kt
@@ -9,7 +9,9 @@ package it.niedermann.owncloud.notes.share
 import android.content.Intent
 import android.content.res.Configuration
 import android.os.Bundle
+import android.text.Editable
 import android.text.TextUtils
+import android.text.TextWatcher
 import android.view.View
 import androidx.lifecycle.lifecycleScope
 import com.nextcloud.android.common.ui.theme.utils.ColorRole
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
When I click on three dots menu for an interal share, I see "delete link", but it should be "unshare" (or smth like that)

tobiasKaminsky:
Can you retest this, on web:
- user2 shares a note to user1
- it ends up in /Shared for user1
- not visible in Notes for user1

If yes, this is a bug on Notes web app

alperozturk96:
@tobiasKaminsky @AndyScherzinger  force merge needed

PR #3192, opened 2026-05-16
fix(direct-edit): route off-server links out of the editor WebView

### LINKED ISSUE #3191: NoteDirectEditFragment WebView navigates itself for off-server links
`NoteDirectEditFragment` (`app/src/main/java/it/niedermann/owncloud/notes/edit/NoteDirectEditFragment.kt`) loads the Nextcloud Text editor for the user's account and configures the `WebViewClient` with `onReceivedError` and `onReceivedSslError` overrides but no `shouldOverrideUrlLoading`.

When the rendered editor contains links (a markdown link to another site, a mention to a user on a different Nextcloud server, a `mailto:` in a frontmatter table) and the user taps one, the WebView navigates itself to that URL. The edit session is lost and the user is now stuck in a non-editor page inside a fragment that was only set up to render the editor.

Most Nextcloud Android surfaces (Files, Talk, ...) handle this by keeping same-server URLs inside the WebView and routing everything else out via `Intent.ACTION_VIEW`.

### Suggested fix

Override `shouldOverrideUrlLoading`:
1. If the URL starts with the current account URL, return `false` so it stays in the WebView.
2. Otherwise, dispatch `Intent.ACTION_VIEW` and wrap the launch in `try/catch (ActivityNotFoundException)` so a device with no handler for the scheme does not crash the fragment.

```kotlin
override fun shouldOverrideUrlLoading(view: WebView?, request: WebResourceRequest?): Boolean {
    val url = request?.url ?: return false
    val scheme = url.scheme?.lowercase()
    val accountUrl = runCatching { account.url }.getOrNull()
    if ((scheme == "http" || scheme == "https") &&
        accountUrl != null && url.toString().startsWith(accountUrl)
    ) {
        return false
    }
    return try {
        val intent = Intent(Intent.ACTION_VIEW, url).apply {
            addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
        }
        startActivity(intent)
        true
    } catch (e: ActivityNotFoundException) {
        Log.w(TAG, "No app available to open $url", e)
        true
    }
}
```

A PR with the change is at #3192.

jim-daf:
Closes #3191.

The `NoteDirectEditFragment` WebView loads the Nextcloud Text editor for the user's account but does not override `shouldOverrideUrlLoading`. Any link the editor renders (a markdown link to another site, a `mailto:` in a frontmatter table, a mention to a user on a different Nextcloud server) navigates the editor WebView itself, which drops the user out of the edit session.

### Change

Override `shouldOverrideUrlLoading`. Keep navigations whose URL starts with the current account URL inside the WebView (so server-internal links the editor produces continue to work). Route everything else out via `Intent.ACTION_VIEW`. Wrap the launch in `try/catch (ActivityNotFoundException)` so a missing handler does not crash the Fragment:

```kotlin
override fun shouldOverrideUrlLoading(view: WebView?, request: WebResourceRequest?): Boolean {
    val url = request?.url ?: return false
    val scheme = url.scheme?.lowercase()
    val accountUrl = runCatching { account.url }.getOrNull()
    if ((scheme == "http" || scheme == "https") &&
        accountUrl != null && url.toString().startsWith(accountUrl)
    ) {
        return false
    }
    return try {
        val intent = Intent(Intent.ACTION_VIEW, url).apply {
            addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
        }
        startActivity(intent)
        true
    } catch (e: ActivityNotFoundException) {
        Log.w(TAG, "No app available to open $url", e)
        true
    }
}
```

The `runCatching` around `account.url` is so a transient access failure here cannot crash the WebViewClient callback. If the account URL is not available we just fall through to the external-route branch instead of blowing up.

### Behaviour

- Same-server `https` link inside the editor (the case the editor actually produces): unchanged, stays in the WebView.
- External `https` link tapped from a note: opens in the system browser.
- `mailto:`, `tel:`, `intent:` link tapped from a note: opens in the matching system app.
- No handler installed: warning logged, no crash.


tobiasKaminsky (approved):

tobiasKaminsky:
Nice idea!

Can you check Codacy and sign DCO: 
https://github.com/nextcloud/notes-android/pull/3192/checks?check_run_id=76320050514

jim-daf:
> Nice idea!
> 
> Can you check Codacy and sign DCO: https://github.com/nextcloud/notes-android/pull/3192/checks?check_run_id=76320050514

Hi @tobiasKaminsky i checked both Codacy and DCO

tobiasKaminsky:
Thank you very much :tada:

PR #3203, opened 2026-05-27
feat: remember preview/edit mode per note

jancborchardt:
In Nextcloud Notes for Android, currently the Preview/Edit toggle in the top right of the note detail view changes this setting for all notes. It would be nice to have the setting remembered implicitly per note.

This is nice if you have some notes which are usually more useful in view-mode, like a packlist or grocery list where you want to check off items.

## Summary

- Adds a `noteMode` column to the `Note` table (DB migration 29 → 30) to persist each note's view mode locally
- When a user explicitly toggles Preview/Edit/Direct Edit, the selected mode is saved immediately to that note
- Notes that have never been toggled continue to follow the global preference (Settings → Note display mode) unchanged
- Also commits the missing `29.json` schema file required for the existing `@AutoMigration(from = 28, to = 29)`

Resolves the existing TODO comment in `EditNoteActivity.close()`:
> TODO enhancement: store last mode in note

## How it works

`getPreferenceMode()` now checks the note's stored `noteMode` first, and only falls back to the global SharedPreference if no per-note mode is set. `changeMode()` persists the new mode to the note immediately on toggle.

## Test plan

- [ ] Open a note → toggle to Preview → back out → re-open: should open in Preview
- [ ] Open a different note that was never toggled: should open in the global default mode
- [ ] Change Settings → Note display mode: notes with no per-note mode should follow the new setting; notes with a stored mode should not
- [ ] Toggle a note to Edit → close → re-open: should open in Edit
- [ ] Upgrade from a previous install: existing notes should open in the global default (no stored mode), new toggles are remembered

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/schemas/it.niedermann.owncloud.notes.persistence.NotesDatabase/29.json
+++ b/app/schemas/it.niedermann.owncloud.notes.persistence.NotesDatabase/29.json
@@ -1,0 +1,756 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 29,
+    "identityHash": "8945d8fc9edbc1283151e51c2e201541",
+    "entities": [
+      {
+        "tableName": "Account",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`id` INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, `url` TEXT NOT NULL DEFAULT '', `userName` TEXT NOT NULL DEFAULT '', `accountName` TEXT NOT NULL DEFAULT '', `eTag` TEXT, `modified` INTEGER, `apiVersion` TEXT, `color` INTEGER NOT NULL DEFAULT -16743735, `textColor` INTEGER NOT NULL DEFAULT -16777216, `capabilitiesETag` TEXT, `displayName` TEXT, `directEditingAvailable` INTEGER NOT NULL)",
+        "fields": [
+          {
+            "fieldPath": "id",
+            "columnName": "id",
+            "affinity": "INTEGER",
+            "notNull": true
+          },
+          {
+            "fieldPath": "url",
+            "columnName": "url",
+            "affinity": "TEXT",
+            "notNull": true,
+            "defaultValue": "''"
+          },
+          {
+            "fieldPath": "userName",
+            "columnName": "userName",
+            "affinity": "TEXT",
+            "notNull": true,
+            "defaultValue": "''"
+          },
+          {
+            "fieldPath": "accountName",
+            "columnName": "accountName",
+            "affinity": "TEXT",
+            "notNull": true,
+            "defaultValue": "''"
+          },
+          {
+            "fieldPath": "eTag",
+            "columnName": "eTag",
+            "affinity": "TEXT"
+          },
+          {
+            "fieldPath": "modified",
+            "columnName": "modified",
+            "affinity": "INTEGER"
+          },
+          {
+            "fieldPath": "apiVersion",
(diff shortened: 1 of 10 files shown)


app/src/main/java/it/niedermann/owncloud/notes/persistence/migration/Migration_29_30.kt
```diff
@@ -0,0 +1,17 @@
...
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+package it.niedermann.owncloud.notes.persistence.migration
+
+import androidx.room.migration.Migration
+import androidx.sqlite.db.SupportSQLiteDatabase
+
+@Suppress("ClassName", "Detekt.ClassNaming")
+class Migration_29_30 : Migration(29, 30) {
```
alperozturk96: Auto migration should be enough to add new column to the note table

app/src/main/java/it/niedermann/owncloud/notes/edit/EditNoteActivity.java
```diff
@@ -249,17 +249,31 @@ private void replaceFragment() {
...
+     * Returns the preferred mode for the note. Checks the per-note stored mode first, then falls
+     * back to the global preference. If the mode is "remember last" the last mode is returned.
+     * If the mode is "direct edit" and the account does not support direct edit, edit mode is returned.
      */
-    private String getPreferenceMode(long accountId) {
+    private String getPreferenceMode(long accountId, long noteId) {
+        final var defaultMode = getString(R.string.pref_value_mode_edit);
+        final var prefValueDirectEdit = getString(R.string.pref_value_mode_direct_edit);
+
+        if (noteId > 0) {
```
alperozturk96: Please use early returns

app/schemas/it.niedermann.owncloud.notes.persistence.NotesDatabase/29.json
```diff
@@ -0,0 +1,756 @@
+{
```
alperozturk96, line 1: Seems like 29.json never added before @tobiasKaminsky ?

alperozturk96 (changes requested):

tobiasKaminsky:
How will this work with? 
image


As far as I can see from code side, it would always use per-note setting?
But it should only keep it if last one is selected? 
Or do we wanna change this to "First opening behaviour" (and remove "remember").

Then user can select what they want to do with new notes, and all other notes remember their last behaviour?

app/src/main/java/it/niedermann/owncloud/notes/persistence/migration/Migration_29_30.kt
jancborchardt: Resolved

app/src/main/java/it/niedermann/owncloud/notes/edit/EditNoteActivity.java
jancborchardt: Resolved

jancborchardt:
@tobiasKaminsky yeah, makes sense, adjusted and removed the remember option, as it always remembers. Evveryone who had "Remember last" selected now defaults to "Edit" for untouched notes and their per-note toggles are remembered going forward.

@alperozturk96 your review comments are also addressed. :)

alperozturk96 (approved):
Tested working.

PR #3215, opened 2026-06-04
feat(widget): add interactive notes widget

loloakira:
## Description

This PR adds a new home-screen widget called **Interactive notes** — a third widget alongside the existing single-note and note-list widgets, which are left unchanged.

### What it does
- Shows notes as a scrollable single column of rounded cards.
- Each card shows the note title and a multi-line content excerpt (the whole note when it is short).
- Tapping a card opens that specific note; a floating "+" button in the bottom-right creates a new note in the widget's configured category/account.
- The favorite star is shown as a non-clickable indicator.
- Reconfigurable via long-press on Android 12+, with a live previewLayout in the widget picker.

### Implementation notes
- Reuses the existing `NotesListWidgetData` storage and excerpt generation - **no database migration**.
- Sort options are stored in per-widget `SharedPreferences`; notes are sorted in memory.
- One shared change: WidgetNotesListDao.createOrUpdateNoteListWidgetData now uses @Insert(onConflict = OnConflictStrategy.REPLACE) instead of a plain insert that aborted on the existing primary key, so reconfiguration can persist updated settings. This is transparent to the two existing widgets, which are not reconfigurable and never hit the conflict. A DAO test covers the upsert.
- New source strings were added to `values/strings.xml` only (translations are handled by Transifex).
- The two existing widgets and the shared data layer are untouched.
- Added the explicit "Add widget" button so that saving is no longer triggered by tapping a category.
- Added section headings and preselected "All notes".

### Tested
- `./gradlew assembleDevDebug` and `./gradlew lintDevDebug` pass.
- test DAO + `assembleDevDebug/lintDevDebug` pass
- Manually verified on a device (Android 16): open note, create note, sort options, reconfiguration, light/dark mode.
  
### 🖼️Screenshots

🏚 Before | 🏡 After
---|---   
No equivalent widget | interactive_note_list_widget_preview

### 🏁 Checklist
  
- [x] 🧪 Tests written - WidgetNotesListDaoTest covers the DAO upsert; the UI was verified manually.
  
---
Developed with AI assistance (Claude Code 2.1.219, Claude Opus 4.8 for the initial implementation and Claude Opus 5 for the review changes)


tobiasKaminsky:
Awesome!

Just a smal UX thing:
When I first set it up, I was unsure how to "save".
I had to click on "all notes" (or Favorites).
I am a bit unsure how to make this more understandable?
Maybe via a radio button and a real "save/add" button?

@jancborchardt do you have an idea?

image

loloakira:
Thanks!

I don't like it either but as it was my first PR for this project I didn't want to change too much things. I reused the interface of the previous widget that wasn't editable but where the saving was done on selecting the category.

tobiasKaminsky:
I understand. 
Let's see if @jancborchardt has a good idea, which does not require too many changes.

jancborchardt (commented):
Looks very nice, and useful! :)

Regarding the settings, some details:
- Settings regarding favorites on top and sorting could just follow in-app preferences. Either they can be used as a preset here, or we don't need to show them at all here
- "All notes" could be preselected, and yes I would say a specific FAB in the bottom right saying "Add widget" would be nice (needs to be ("Edit widget" when editing of course)

loloakira:
Thank you!

I had a little time to make those changes and upgrade the project to the last main commit. 

- I fixed the minSdk part in my widget and modified the saving menu. 
- I added a FAB. 
- I added titles to understand the different parts and made "All notes" selected by default. It's now a selection before saving the widget with the FAB and not a saving on selecting the category. 
- I checked the default configuration parts but found no place where the the newest/oldest and favorites on top are so I just let favorites on by default like the rest of the app. 
- I used the US form and not UK form for "Favorites" instead of "Favourites". 

Now I like this menu. Tell me if it's ok.

app/src/main/java/it/niedermann/owncloud/notes/widget/interactivelist/InteractiveNoteListWidget.kt
```diff
@@ -0,0 +1,179 @@
...
+            Log.w(TAG, "Intent action is null")
+            return
+        }
+
+        if (intent.action != AppWidgetManager.ACTION_APPWIDGET_UPDATE) {
+            Log.w(TAG, "Intent action is not ACTION_APPWIDGET_UPDATE")
+            return
+        }
+
+        if (!intent.hasExtra(AppWidgetManager.EXTRA_APPWIDGET_ID)) {
```
alperozturk96: If we don't have this `EXTRA_APPWIDGET_ID` we update widget with 

` awm.getAppWidgetIds(ComponentName(context, InteractiveNoteListWidget::class.java))`

Then still update widget with `-1`.

Why this approach is preferred? Shouldn't we use one of it only instead of both.

app/src/main/java/it/niedermann/owncloud/notes/widget/interactivelist/InteractiveNoteListWidgetConfigurationActivity.java
```diff
@@ -0,0 +1,239 @@
+/*
+ * Nextcloud Notes - Android Client
+ *
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+package it.niedermann.owncloud.notes.widget.interactivelist;
```
alperozturk96: Can we use `Kotlin` for new class?

app/src/main/java/it/niedermann/owncloud/notes/widget/interactivelist/InteractiveNoteListWidgetService.java
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+package it.niedermann.owncloud.notes.widget.interactivelist;
+
+import android.content.Intent;
+import android.widget.RemoteViewsService;
+
+public class InteractiveNoteListWidgetService extends RemoteViewsService {
```
alperozturk96: Can we use `Kotlin` for new class?

app/src/main/java/it/niedermann/owncloud/notes/widget/interactivelist/InteractiveWidgetPreferences.kt
```diff
@@ -0,0 +1,48 @@
...
+
+    @JvmStatic
+    fun save(context: Context, appWidgetId: Int, favoritesFirst: Boolean, sortOrder: WidgetSortOrder) {
+        prefs(context).edit {
+            putBoolean(KEY_FAVORITES_FIRST + appWidgetId, favoritesFirst)
+            putString(KEY_SORT_ORDER + appWidgetId, sortOrder.name)
+        }
+    }
+
+    @JvmStatic
```
alperozturk96: No need `@JvmStatic` if we use Kotlin for other new classes.

app/src/main/java/it/niedermann/owncloud/notes/widget/interactivelist/InteractiveWidgetSelection.kt
```diff
@@ -0,0 +1,36 @@
...
+package it.niedermann.owncloud.notes.widget.interactivelist
+
+import it.niedermann.owncloud.notes.main.MainActivity
+import it.niedermann.owncloud.notes.persistence.entity.NotesListWidgetData
+
+object InteractiveWidgetSelection {
+    private const val CATEGORY_ID_PREFIX = "category:"
+    private const val CATEGORY_SEPARATOR = '/'
+
+    @JvmStatic
```
alperozturk96: No need @JvmStatic if we use Kotlin for other new classes.

app/src/main/res/drawable/interactive_note_list_widget_preview.webp
alperozturk96, line 1: This preview should be deleted from repo.

alperozturk96 (changes requested):

app/src/main/res/drawable/interactive_note_list_widget_preview.webp
loloakira: I don't think so because minSdk is 28 and previewLayout is API 31. Android 9, 10 and 11 only read previewImage so without the icon, we would have the app icon as preview. I initially added this screenshot because note_list_widget_preview.webp and single_note_widget.webp were in res/drawable/ and listed in REUSE.toml so I just copied previous widget formats. 

If you're sure it must go, I just have to delete it, it's the last change I haven't done yet.

app/src/main/java/it/niedermann/owncloud/notes/widget/interactivelist/InteractiveNoteListWidgetConfigurationActivity.java
loloakira: There were no instruction on new classes when I did the PR and old widgets were in java. As I said, I didn't wanted to make multiple changes on my first PR on this project but I just made the changes.

app/src/main/java/it/niedermann/owncloud/notes/widget/interactivelist/InteractiveNoteListWidget.kt
loloakira: That was a copy of the existing NoteListWidget.kt but I made the fix on my widget. I didn't touch other widgets.

app/src/main/res/drawable/interactive_note_list_widget_preview.webp
alperozturk96: Please delete the preview. Repository is not keeping previews for each new features.

alperozturk96 (changes requested):

alperozturk96 (changes requested):
Widget not shows the content like preview. Content is different.

<table>
  <tr>
    <th>Note preview</th>
    <th>Widget</th>
  </tr>
  <tr>
    <td>before</td>
    <td>after</td>
  </tr>
</table>

app/src/main/res/drawable/interactive_note_list_widget_preview.webp
loloakira: It's deleted

loloakira:
The problem shown in your screenshot is not part of the widget. The bug lies in NoteUtil.generateNoteExcerpt and the title was removed with a string check on the content. I didn't see it because my notes didn't include the title in the first line but in your screenshot "Asdasd" is the title and "Asdasd1" is your first line so when the title is stripped, it only leaves the "1". 
I made a fix in a second commit on my branch but as excerpts are persisted, notes written before this change keep their truncated excerpt. 
As the last parts starts to touch parts I didn't plan to initially touch, I don't know how to follow up. Keep me posted if I can be of any further help.

Issue #3162, opened 2026-04-24
Typing in long md files is slow (~1s delay)


Lenzelot:
### This issue respects the following points:

- [x] I have read the [FAQ](https://github.com/nextcloud/notes-android/blob/master/FAQ.md)
- [x] I have searched for existing issues
- [x] The issue affects the Notes Android app, neither the [Notes server app](https://github.com/nextcloud/notes/issues/) nor the [Nextcloud Android app](https://github.com/nextcloud/android/issues).

### Describe the bug

First of all, sorry in advance if this issue is "wrong" here. I did a check and this affects the Markdown web editor and the editor in the Notes server app aswell. Do with it as you please.

Got a couple of very long notes files. 5000 lines or more... (please don't judge). 

- 5000 lines or more in one note markdown file
- App is in Markdown code view
- Typing anywhere in the file with the android keyboard

Typing in them has like a one second delay on each keystroke.  I'm no expert but is there some very inefficient processing going on? Like, is it going through the entire file each time you change something? Because this delay is cumulative. 5 keystrokes =  roughly 5 times the time. I have tested a couple of different markdown editors and while this very issue is certainly not unique, there are other apps that behave normally, which is no delay at all. 

If you think this kind of delay is normal or acceptable, I don't know what to tell you. It's a plain text file and there are ways to deal with syntax highlighting efficiently. All that being said, I appreciate developers efforts here, all in all the app works great and gets the job done.

### Notes Android version

33.0.0

### Notes server version

33.0.x

### Nextcloud Android version

33.0.1

### Nextcloud version

33.0.x

### Device

Pixel 8

### Android Version

16

### App Store

- [ ] Google Play Store
- [x] F-Droid
- [ ] Huawei App Gallery

MiMoHo:
Likely the same root cause as #2489 (per-keystroke thread pool creation in
`nextcloud-commons:markdown` since the 2.x controller framework, plus full-document
re-rendering). Proposed fix: stefan-niedermann/nextcloud-commons#436.

Issue #2489, opened 2024-11-15
Extreme Performance Issues With Long Notes


solonovamax:
### This issue respects the following points:

- [x] I have read the [FAQ](https://github.com/nextcloud/notes-android/blob/master/FAQ.md)
- [x] I have searched for existing issues
- [x] The issue affects the Notes Android app, neither the [Notes server app](https://github.com/nextcloud/notes/issues/) nor the [Nextcloud Android app](https://github.com/nextcloud/android/issues).

### Describe the bug

Editing long notes causes extreme performance issues, where there is noticeable input lag when typing.

This issue only occurs when using plain editing mode and not the rich edit mode. However, using the plain edit mode is a requirement for me, as I can often times experience connection drop outs.

This is a regression, as previous versions of the notes app did not experience this. Later, I will attempt to determine which version introduced the regression, however I believe it was introduced somewhere in the range of 6-9 months ago (I never bothered to report it at that time).

To reproduce:
1. Create a new note
2. Switch to plain edit mode
3. Copy the below text into the note
4. Attempt to type anything

See attached video

https://github.com/user-attachments/assets/b3bc5ea0-c508-4d1d-b764-1680ebea30c5

### Expected behavior

Does not have performance issues with long notes

### Notes Android version

4.3.1

### Notes server version

4.11.0

### Nextcloud Android version

3.30.4

### Nextcloud version

29.0.7

### Device

OnePlus 9 Pro 5G

### Android Version

14

### App Store

- [ ] Google Play Store
- [x] F-Droid
- [ ] Huawei App Gallery

joshtrichards:
How large is this note file? (I.e. Check it in the Web UI of the *Files* app in your *Notes* folder). This will help us test a similar size.

From your video it looks like not tiny, but also not an incredibly large note (judging by what I can guess/estimate based the scrollbar size).

I know slowdowns happen with very large notes (this app is optimized for *notes* not massive documents), but notes can still get fairly large before that happens (speaking solely in terms of my own experience).

I wonder if it's specific to resource availability on the device. That might be a variable too.

solonovamax:
> How large is this note file? (I.e. Check it in the Web UI of the _Files_ app in your _Notes_ folder). This will help us test a similar size.
> 
> From your video it looks like not tiny, but also not an incredibly large note (judging by what I can guess/estimate based the scrollbar size).

The size of the note is 81 kb, according to the Files web UI.
According to `wc`, here are some additional metrics:
| | |
[...]

PR #3249, opened 2026-07-02
perf(search): adaptive debounce and background matching for in-note search

### LINKED ISSUE #1729: Slow search in long notes

MiMoHo:
## Problem

Fixes #1729. In long notes, in-note search starts searching after the very first
typed letter. Counting, nth-occurrence lookup and highlighting all ran synchronously
on the main thread, so a single-letter query with thousands of matches freezes the
UI for seconds (see the "&"-prefix workaround described in #1729).

## Changes

1. **Adaptive debounce** instead of the static 50/200 ms delays: the delay scales
   with the *measured* duration of the previous search + highlight pass —
   `clamp(2 × lastDuration, 50 ms, 750 ms)`, plus 150 ms for queries ≤ 3 characters.
   Short notes keep the current snappy behavior (addressing the earlier concern in
   #1729 that a fixed 1 s delay would hurt every use case), while long notes
   automatically get a delay long enough to let the user finish typing. Measuring
   real duration is more robust than scaling by content length, because the cost is
   dominated by the number of matches.
2. **Background matching with cancellation**: occurrence counting runs on a
   background executor; a generation counter discards superseded results ("some kind
   of Cancel event", as suggested by the maintainer in #1729). Only the newest query
   is highlighted, on the main thread.
3. **No more full copies per query**: `jumpToOccurrence()` allocated a full
   `toLowerCase()` copy plus a `substring(0, index)` copy of the entire note per
   query. Replaced by a single case-insensitive `Matcher` pass; the recursive,
   case-sensitive `indexOfNth` is gone (it also disagreed with the case-insensitive
   counting/highlighting).

## Not covered

The remaining per-keystroke cost of the editor itself lives in
`nextcloud-commons:markdown` — see the companion PR
stefan-niedermann/nextcloud-commons#436 (executor reuse in
`EditorStateNotifier`, debounced `SearchHighlightTextWatcher`). Bumping
`commonsVersion` after its release will address #2489 / #3162.

## Testing

- Existing unit tests pass (`./gradlew test`).
- Manual: note with ~100 kB text; type a common letter ("e") in the search bar and
  continue typing — the UI no longer freezes after the first letter; searching in a
  short note behaves as before.

### DIFF
--- a/app/src/main/java/it/niedermann/owncloud/notes/edit/SearchableBaseNoteFragment.java
+++ b/app/src/main/java/it/niedermann/owncloud/notes/edit/SearchableBaseNoteFragment.java
@@ -8,6 +8,8 @@ package it.niedermann.owncloud.notes.edit;
 
 import android.os.Bundle;
 import android.os.Handler;
+import android.os.Looper;
+import android.os.SystemClock;
 import android.text.Layout;
 import android.text.TextUtils;
 import android.util.Log;
@@ -30,6 +32,9 @@ import com.nextcloud.android.sso.exceptions.NoCurrentAccountSelectedException;
 import com.nextcloud.android.sso.helper.SingleAccountHelper;
 import com.nextcloud.android.sso.model.SingleSignOnAccount;
 
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.regex.Pattern;
 
 import it.niedermann.owncloud.notes.R;
@@ -47,9 +52,14 @@ public abstract class SearchableBaseNoteFragment extends BaseNoteFragment {
     private int occurrenceCount = 0;
     private SearchView searchView;
     private String searchQuery = null;
-    private static final int delay = 50; // If the search string does not change after $delay ms, then the search task starts.
-    private static final int shortStringDelay = 200; // A longer delay for short search strings.
+    private static final int minDelay = 50; // Minimum delay in ms after the search string stopped changing before the search task starts.
+    private static final int maxDelay = 750; // Upper bound in ms for the adaptive search delay.
+    private static final int shortStringExtraDelay = 150; // Additional delay for short search strings because they tend to produce many matches and therefore expensive highlighting.
     private static final int shortStringSize = 3; // The maximum length of a short search string.
+    private long lastSearchDuration = 0; // Measured duration of the last search + highlight pass. Used to adapt the debounce delay to the size of the note and the amount of matches (#1729).
+    private final AtomicInteger searchGeneration = new AtomicInteger();
+    private final ExecutorService searchExecutor = Executors.newSingleThreadExecutor();
+    private final Handler mainHandler = new Handler(Looper.getMainLooper());
     private boolean directEditRemotelyAvailable = false; // avoid using this directly, instead use: isDirectEditEnabled()
 
     @ColorInt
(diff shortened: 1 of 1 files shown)


app/src/main/java/it/niedermann/owncloud/notes/edit/SearchableBaseNoteFragment.java
```diff
@@ -47,9 +52,14 @@ public abstract class SearchableBaseNoteFragment extends BaseNoteFragment {
     private int occurrenceCount = 0;
     private SearchView searchView;
     private String searchQuery = null;
-    private static final int delay = 50; // If the search string does not change after $delay ms, then the search task starts.
-    private static final int shortStringDelay = 200; // A longer delay for short search strings.
+    private static final int minDelay = 50; // Minimum delay in ms after the search string stopped changing before the search task starts.
```
alperozturk96, line 55: Thank you for the PR. Good idea.

I think we could make the implementation a bit cleaner and easier to maintain. Instead of adding the debounce logic directly to `SearchableBaseNoteFragment.java`, would it make sense to move this into a small Kotlin helper class and use coroutines for the implementation?

For example:

```kotlin
class SearchHelper(private val fragment: Fragment) {
    companion object {
        private const val MAX_DELAY = 750L
        private const val LAST_SEARCH_DURATION = 0L
        // ...
    }

    fun checkOccurrences(onComplete: () -> Unit) {
        fragment.lifecycleOwner(Dispatchers.IO) {
            // ...
            withContext(Dispatchers.Main) {
                onComplete()
            }
        }
    }
    
    // if possible we can add `countOccurrences` here as well.
}
```

This would give us lifecycle-aware code and would also remove the need for managing an `ExecutorService` and calling `shutdownNow()` explicitly.

One other suggestion: I think we can remove the comments and keep the variable/method names self-explanatory.

As we add more fixes and features to the fragment, mixing this logic directly into it could make the file increasingly difficult to follow.

I think the approach is good. Thanks again for the work on this.

alperozturk96 (changes requested):

